PR #175, opened 2025-01-02
refactor: using try with sources and remove duplicate try catch

GAhlin:


psxjoy:
Hi, this PR feels a bit unexpected. Should we consider linking it to an issue or outlining any prerequisites for the proposed optimizations?

PR #244, opened 2025-03-02
chore: drop old code

tisonkun:
These files seem less useful.

Also, the links inside are broken.


tisonkun:
cc @jipengfei-jpf @psxjoy would you give a review and trigger the workflow?

psxjoy:
LGTM

Issue #228, opened 2025-02-17
填充时日期格式丢失问题


721806280:
```
@Getter
@Setter
@EqualsAndHashCode
public class FillData {
    private String name;
    private double number;
    private Date date;
    private LocalDate date2;
}
```

------------------------------------------------------------------------------

```
@Test
    public void t01ReadAndWrite() throws Exception {
        List<FillData> list = ListUtils.newArrayList();
        for (int i = 0; i < 10; i++) {
            FillData fillData = new FillData();
            list.add(fillData);
            fillData.setName("张三");
            fillData.setNumber(5.2);
            fillData.setDate(new Date());
            fillData.setDate2(LocalDate.now());
        }

        EasyExcel.write().file(file07).head(FillData.class)
                .withTemplate(fileTemplate07).sheet().doFill(list);

    }
```

------------------------------------------------------------------------------

[annotation.xlsx](https://github.com/user-attachments/files/18820820/annotation.xlsx)

<img width="904" alt="Image" src="https://github.com/user-attachments/assets/9d199f16-17b0-4aa0-b480-0c1a721acaac" />

@psxjoy 🐛  BUG. 请优先修复

721806280:
@psxjoy 参考 https://github.com/alibaba/easyexcel/pull/3366

psxjoy:
Hi, thanks for raising this issue. This is something we are already planning to address. To maintain compliance with open-source licensing, we are in the process of contacting the relevant PR authors. I am confident that this will be fixed in the upcoming version.

senningsoft:
填充数据时能根据条件改变某行数据的样式吗？如改变字体的颜色

721806280:
@psxjoy Please make a reasonable assessment as to whether the location of the correction is more appropriate here.

https://github.com/fast-excel/fastexcel/blob/370c6029609f1c9b5604fe8841fcabf682017af0/fastexcel-core/src/main/java/cn/idev/excel/write/executor/ExcelWriteFillExecutor.java#L288
```
               // Restyle
                if (fillConfig.getAutoStyle()) {
                    Optional.ofNullable(collectionFieldStyleCache.get(currentUniqueDataFlag))
                        .map(collectionFieldStyleMap -> collectionFieldStyleMap.get(analysisCell))
                        .ifPresent(cellStyle -> {
                            cellData.setOriginCellStyle(cellStyle);
[...]

PR #273, opened 2025-03-19
fix: 填充时日期格式丢失问题

waterisblue:
测试代码如下：
```java
@Setter
@Getter
static class FillData {
    private String name;
    private double number;
    private Date date;
    private LocalDate date2;
}
@Test
public void templateWriteFix() {
    String templateFileName = TestFileUtil.getPath() + "demo" + File.separator + "annotation.xlsx";
    String fileName = TestFileUtil.getPath() + "templateWrite" + System.currentTimeMillis() + ".xlsx";
    
    List<FillData> list = ListUtils.newArrayList();
    for (int i = 0; i < 10; i++) {
        FillData fillData = new FillData();
        list.add(fillData);
        fillData.setName("张三");
        fillData.setNumber(5.21111);
        fillData.setDate(new Date());
        fillData.setDate2(LocalDate.now());
    }

    EasyExcel.write(fileName, FillData.class).withTemplate(templateFileName).sheet().doFill(list);
}
```
#228


psxjoy (changes requested):
Please add unit-test code cases in `test` module.

waterisblue:
When the date cell format is set:
image
Output:
image
When the date cell format is default:
image
Output:
image

721806280:
Please make a reasonable assessment as to whether the location of the correction is more appropriate here.
https://github.com/fast-excel/fastexcel/issues/228#issuecomment-2735994781

waterisblue:
Yes, this solution has minor impact on the project and can address #228 , but IMO:
1. Modifying the format of a specific type in the `doFill` method may be somewhat inconsistent with the programming standard. I believe it is more appropriate to modify the styles in the `createCellStyle` method.
2. Discarding type checks and directly calling `DateUtils.isADateFormat` may affect performance, maybe no fast, of course, this requires more testing.
@721806280 @psxjoy

psxjoy (approved):

psxjoy:
Both changes work fine. I’ll merge them into master temporarily, and I’ll check the performance details when I have time.
我觉得两个改动都可以。暂时合并到master，性能方面我找时间看下。

PR #283, opened 2025-03-22
fix: failure unit test

alaahong:
Seems some unit tests are based on OS based absolute path, 
just fix them and add mandatory resources


fastexcel-test/src/test/java/cn/idev/excel/test/temp/WriteV33Test.java
```diff
@@ -1,28 +1,23 @@
...
 import cn.idev.excel.util.BooleanUtils;
 import cn.idev.excel.write.handler.CellWriteHandler;
 import cn.idev.excel.write.handler.context.CellWriteHandlerContext;
-
-import org.apache.poi.ss.usermodel.Cell;
-import org.apache.poi.ss.usermodel.CellStyle;
-import org.apache.poi.ss.usermodel.FillPatternType;
-import org.apache.poi.ss.usermodel.IndexedColors;
-import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.ss.usermodel.*;
```
psxjoy: don't use `*` import .

psxjoy (changes requested):

fastexcel-test/src/test/java/cn/idev/excel/test/temp/WriteV33Test.java
alaahong: Just reset as original value

psxjoy:
LGTM

PR #284, opened 2025-03-22
fix: maven properties and module exception for reflection

alaahong:
1. change java version as default maven and spring format

As no usage of `jdk.version` in whole project, so remove and add new properties as traditional format

```
<java.version>1.8</java.version>
<maven.compiler.target>1.8</maven.compiler.target> 
<maven.compiler.source>1.8</maven.compiler.source>
```

reference:
#https://www.baeldung.com/maven-java-version

2. fix the issue while working above java 9+ as java module mechanism limited. Just find out same behavior in github action confugration, so set below as comment in post commit

Add --add-opens for specific package into maven-surefire-plugin

```
<configuration>
    <argLine>
        --add-opens=java.base/java.lang=ALL-UNNAMED 
        --add-opens=java.base/java.util=ALL-UNNAMED 
        --add-opens=java.base/sun.reflect.annotation=ALL-UNNAMED
    </argLine>
</configuration>
```
e.g.
perform the unit test including reflection under java 21

`mvn -Dtest=CacheDataTest test  -D maven.test.skip=false`

Before the change:
image

After the change:
image

reference:
--add-opens module/package=target-module(,target-module)* Updates module to open package to target-module, regardless of module declaration. 
#https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html

### DIFF
--- a/fastexcel-support/pom.xml
+++ b/fastexcel-support/pom.xml
@@ -16,7 +16,6 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <jdk.version>1.8</jdk.version>
         <gpg.skip>true</gpg.skip>
         <maven.javadoc.skip>true</maven.javadoc.skip>
         <maven.test.skip>true</maven.test.skip>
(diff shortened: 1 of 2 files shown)


psxjoy:
LGTM

PR #315, opened 2025-04-12
fix: fix repeat cases

alaahong:
Provide valid repeat file;
Add repeat case for xlsx

Before:
image

After:
image


psxjoy:
Hi, @alaahong can you solve these conflicts file?

psxjoy:
LGTM

PR #348, opened 2025-04-30
fix: fix poi write

alaahong:
1. Format as current standards
2. Replace invalid file
3. Fix resource holding issue
4. Remove useless case

Before:
image

After:
image


psxjoy:
LGTM

PR #370, opened 2025-06-02
dependency: update dependencies and remove exclusions

delei:
IMO,after upgraded POI version to 5.4.1,we can deal with dependency conflicts without using exclusions

## What's changed?

- **commons-csv**: upgrade from 1.11.0 to 1.14.0
- **commons-io**: upgrade from 2.16.1 to 2.18.0
- remove exclusions for commons-io and commons-codec from poi and poi-ooxml dependencies
- update poi and poi-ooxml exclusions in root pom.xml

### DIFF
--- a/fastexcel-core/pom.xml
+++ b/fastexcel-core/pom.xml
@@ -18,46 +18,22 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-csv</artifactId>
         </dependency>
-
         <dependency>
             <groupId>org.apache.poi</groupId>
             <artifactId>poi</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-io</groupId>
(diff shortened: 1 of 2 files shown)


alaahong:
why remove these exclusions? 
seems no need unbind them then involve legacy (not latest version) dependencies.

psxjoy:
same question with @alaahong

delei:
hi,maybe I'm thinking too simple.
When I was dealing with `commons-csv-1.14.0` and `fastexcel-1.2.0` dependency conflicts in my own project，an exception was thrown while reading the CSV file：
`Caused by: java.lang.NoClassDefFoundError: org/apache/commons/io/input/UnsynchronizedBufferedReader`
I just remove these exclusions，I will continue to work on this in my own project

PR #371, opened 2025-06-03
fix: Fix style test

alaahong:
1. remove the case no relationship with FastExcel
2. remove no test case
3. Provide runnable case for style
Before the change
image

Post the change
image


psxjoy (approved):
LGTM

PR #418, opened 2025-07-17
style: update apache license header

delei:

I noticed that many files may have been copied from the other Apache projects,due to historical reasons, these files use the old ASF header comments,and these classes may not be automatically processed by plugins.

## What's changed?

- update apache license header
- reformat these file

## Checklist


psxjoy:
Hi @delei ,I think it's not the core task in this version. Maybe we can do it next time?

delei:
> Hi @delei ,I think it's not the core task in this version. Maybe we can do it next time?


Ok👌

Issue #345, opened 2025-04-28
Module structure


tisonkun:
Hi @psxjoy @jipengfei-jpf!

I noticed that our current module structure is a bit confusing. It has:

* fastexcel: almost an alias of fastexcel-core?
* fastexcel-core: the core interfaces and implementations
* fastexcel-support: bundle spring-core's cglib
* fastexcel-test: testcode and examples

I'd suggest we have:

* (fastexcel or fastexcel-core, choose one of them): the core interfaces and implementations, merge fastexcel-test's testing code into here
* fastexcel-examples (and submodules): holding examples now in fastexcel-test, exclude all these modules when releasing

For fastexcel-support, we can leave it as is for now. But there should be someway to package it properly externally, like:

* https://github.com/apache/flink-shaded
* https://github.com/apache/ratis-thirdparty

The current solution has some issues like #332. I may look into the bundle things later, but the module structure things first.

delei:
完全同意，甚至`fastexcel-examples`，可以单独成一个独立的仓库，用于丰富文档所需的案例，`fastexcel-test`应该只包含`单元测试`所需的代码

Totally agree, even `fastexcel-examples`, could be a separate repository to enrich the cases needed for documentation, and `fastexcel-test` should only contain the code for `unit-tests`

psxjoy:
很好的建议！
-  `fastexcel` 和 `fastexcel-core` 可以合并成一个模块。
- 就像 @delei  提到的那样， fastexcel-examples 可以独立出来，或者作为一个单独的模块，专注于为使用者提供丰富的使用demo.

我认为我们可以继续推进这个改进。
任何人有其他的建议也可以在此讨论。

Very good advice!
- `fastexcel` and `fastexcel-core` could be merged into one module .
- Like @delei mentioned, `fastexcel-examples` could stand independently or be a separate module focused on providing users with rich usage demos.

I think we can move forward with this improvement.
Anyone who has other suggestions can discuss them here as well.

tisonkun:
> fastexcel and fastexcel-core could be merged into one module .

We'd decide one name as the result. `fastexcel`?

psxjoy:
> > fastexcel and fastexcel-core could be merged into one module .
> 
> We'd decide one name as the result. `fastexcel`?

I think `fastexcel` is fine.
Are you already doing this part of the optimization? If it's already in developing state, I can assign this issue to you for now :)

delei:
## Suggestions

[...]

PR #426, opened 2025-07-21
refactor!: merge fastexcel-core into fastexcel

tisonkun:
This is part of #345 

cc @psxjoy @jipengfei-jpf


pom.xml
```diff
@@ -18,7 +18,6 @@
     <inceptionYear>2024</inceptionYear>
 
     <modules>
-        <module>fastexcel-core</module>
         <module>fastexcel-support</module>
         <module>fastexcel-test</module>
         <module>fastexcel</module>
```
alaahong: even the module sequence is not impacting the dependency reference, which means it will pick up fastexcel firstly when building the fastexcel-test, but it also try to build support and test in aggregate build.
how about to set the module as actual compile sequence.

```
        <module>fastexcel</module>
        <module>fastexcel-support</module>
        <module>fastexcel-test</module>
```

tisonkun: Reasonable to include. Updating ...

tisonkun:
```suggestion
        <module>fastexcel</module>
        <module>fastexcel-support</module>
        <module>fastexcel-test</module>
```

psxjoy (approved):
+LGTM

psxjoy:
Thanks @tisonkun , I've merged this PR. let's move forward with what's next :)

Issue #433, opened 2025-07-24
[Bug] Unable to set use1904windowing when reading xlsx files


delei:
### Search before asking

- [x] I searched in the [issues](https://github.com/fast-excel/fastexcel/issues) and found nothing similar.

### Fastexcel version

1.2.0

### JDK version

OpenJDK 1.80_452

## Test case

```java

@Slf4j
public class Issue3745Test {
    @Test
    void testUse1904windowing() throws Exception {
        File testFile = TestFileUtil.createNewFile("issue/easyexcel/test.xlsx");

        // write file
        FastExcel.write(testFile, DemoData.class).excelType(ExcelTypeEnum.XLSX)
                .use1904windowing(Boolean.TRUE)  // set use1904windowing to TRUE
                .sheet()
                .doWrite(demoData(1));

        FastExcel.read(testFile, new AnalysisEventListener<DemoData>() {
                    @Override
                    public void invoke(DemoData data, AnalysisContext context) {
                        boolean isUse1904windowing = context.readSheetHolder().globalConfiguration().getUse1904windowing();
                        log.info("isUse1904windowing: {}", isUse1904windowing);
                        Assertions.assertTrue(isUse1904windowing);
                    }

                    @Override
                    public void doAfterAllAnalysed(AnalysisContext context) {
                    }
                })
                .excelType(ExcelTypeEnum.XLSX)
                .head(DemoData.class)
                .sheet()
                .doReadSync();
    }

    private List<DemoData> demoData(int size) {
        List<DemoData> dataList = new ArrayList<>();
        for (int i = 0; i < size; i++) {
            DemoData demoData = new DemoData();
            demoData.setString("String" + i);
            demoData.setDate(DateUtil.date());
            demoData.setDoubleData(RandomUtil.randomDouble());
            dataList.add(demoData);
        }
        return dataList;
    }
}
```

### What did you expect to see?

For the `xlsx` files, according to the logic of `XlsxSaxAnalyser#analysisUse1904WindowDate` method , the value of `use1904windowing` should be true when reading the file.

### What did you see instead?

The value of `use1904windowing` always be false.

<img width="1786" height="359" alt="Image" src="https://github.com/user-attachments/assets/e48ac186-7576-49cb-a447-64db72b8d127" />

### Related Issue
[easyexcel-issue-3745](https://github.com/alibaba/easyexcel/issues/3745)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

[...]

Issue #434, opened 2025-07-24
[Feature] supports setting setDate1904 when writing to xlsx files


delei:
### Search before asking

- [x] I searched in the [issues](https://github.com/fast-excel/fastexcel/issues) and found nothing similar.

### Motivation

When writing to an xlsx file, the workbook should use the `setDate1904` method and set the value with `use1904windowing`.

In the POI library, the `setDate1904` method is only supported by **XSSFWorkbook** and **SXSSFWorkbook**.

### Solution

The main changes are as follows:
- update the `WorkBookUtil#createWorkBook()` method
- use the `setDate1904` method

```java
// XSSFWorkbook
XSSFWorkbook xssfWorkbook = new XSSFWorkbook();
xssfWorkbook.getCTWorkbook().getWorkbookPr().setDate1904(
    writeWorkbookHolder.getGlobalConfiguration().getUse1904windowing()
);

// SXSSFWorkbook
SXSSFWorkbook sxssfWorkbook = new SXSSFWorkbook();
sxssfWorkbook.getXSSFWorkbook().getCTWorkbook().getWorkbookPr().setDate1904(
    writeWorkbookHolder.getGlobalConfiguration().getUse1904windowing()
);
```

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #435, opened 2025-07-24
feat: add support to set 1904 date system based on global configuration

delei:

Closes: #433 , #434 

## What's changed?
These changes only for the `xlsx` file:
- (write)use the `setDate1904` method to set 1904 date system
- (read)use user-set parameters instead of always using default values.
- (read)add support for handle 1904 date windowing records

## Checklist


fastexcel-test/src/test/java/cn/idev/excel/test/core/parameter/ParameterDataTest.java
```diff
@@ -199,11 +206,51 @@ private void readAndWrite7(File file, ExcelTypeEnum type) {
                 .doRead();
     }
 
+    @Test
+    void readAndWriteDate1904Test() {
+        readAndWriteDate1904(file07, ExcelTypeEnum.XLSX, null);
+        readAndWriteDate1904(file07, ExcelTypeEnum.XLSX, Boolean.TRUE);
+        readAndWriteDate1904(file07, ExcelTypeEnum.XLSX, Boolean.FALSE);
```
alaahong: Just wonder whether XLS always set setWindowing as false in createDateWindow1904()
If so, it's better to write a log to tell user it won't be implemented while set TRUE on use1904windowing for XLS and CSV

delei: When reading `xls` and `csv` files, the value of `use1904windowing` must be set by the user themselves, cannot be automatically obtained from the file.The parameter values obtained are consistent with those set by the user.

delei: I will optimise this unit test.

alaahong: I mean, both CSV and XLS didn't provide the 1904 implementation. CSV is plain text, no relationship with this option, XLS is just like I mentioned, it always setWindowing as false. 
So in your given test case, it will be making user confusing, even they set the TRUE and read TRUE value from configuration, the actual file still be treating as FALSE.

alaahong: As below line mentioned, might be it won't work at this moment.  
https://github.com/apache/poi/blob/bcc7912c8dc284e6f1b345c9295235ecf4fdce27/poi/src/main/java/org/apache/poi/hssf/model/InternalWorkbook.java#L1269

alaahong (changes requested):
Please concern to add full test not only on the configuration value, it should be matching the result with configuration set.

XLS--> https://github.com/apache/poi/blob/bcc7912c8dc284e6f1b345c9295235ecf4fdce27/poi/src/main/java/org/apache/poi/hssf/model/InternalWorkbook.java#L1269

fastexcel-test/src/test/java/cn/idev/excel/test/core/parameter/ParameterDataTest.java
delei: You're right.Even the user set the TRUE and read TRUE value from configuration, the **actual file** still be treating as FALSE.
Since it is not possible to automatically parse from the file, FastExcel uses the values set by the user to convert date-type when reading the CSV and XLS files.

delei:
> [Please concern to add full test not only on the configuration value, it should be matching the result with configuration set.](https://github.com/apache/poi/blob/bcc7912c8dc284e6f1b345c9295235ecf4fdce27/poi/src/main/java/org/apache/poi/hssf/model/InternalWorkbook.java#L1269)


Thank you for the information. I will check to see if it is supported.

delei:
I will refactor the unit tests for datewindowing ASAP.

fastexcel-test/src/test/java/cn/idev/excel/test/core/parameter/DateWindowingTest.java
```diff
@@ -0,0 +1,142 @@
...
+        file03DateWindowing1904 = TestFileUtil.createNewFile("datewindowing1904.xls");
+        file03DateWindowing1904Exists = TestFileUtil.readFile("datewindowing/1904DateWindowing.xls");
+        file07DateWindowing1900 = TestFileUtil.createNewFile("datewindowing1900.xlsx");
+        file07DateWindowing1904 = TestFileUtil.createNewFile("datewindowing1904.xlsx");
+        fileCsvDateWindowing1900 = TestFileUtil.createNewFile("datewindowing1900.csv");
+        fileCsvDateWindowing1904 = TestFileUtil.createNewFile("datewindowing1904.csv");
+    }
+
+    @Test
+    void test03WriteAndRead() {
```
alaahong, line 42: seems no assertion or verify here.  it should be throwing exception, or write warn log to keep no breaking change at least. 
but the test highlight that FastExcel won't support but test pass with a true value.

delei: updated, others assertions in class `DateWindowingListener`

psxjoy (approved):
Sorry for not checking this PR in time. I don't see any issues with it, and once @alaahong  confirms that it's fine, I will merge the PR within 24 hours.

fastexcel-test/src/test/java/cn/idev/excel/test/core/parameter/DateWindowingTest.java
```diff
@@ -0,0 +1,156 @@
...
+                .use1904windowing(Boolean.FALSE)
+                .sheet()
+                .doWrite(data());
+
+        // not support setting actual file to true.
+        FastExcel.write(file03DateWindowing1904, ParameterData.class)
+                .excelType(ExcelTypeEnum.XLS)
+                .use1904windowing(Boolean.TRUE)
+                .sheet()
+                .doWrite(data());
```
alaahong, line 55: may I ask for this setting behavior?  will the excel show the 1904 model since wrote with true?
image

alaahong (commented):
Generally looks good. just leave a question for detail, please kindly help to check if you've a time.

fastexcel-test/src/test/java/cn/idev/excel/test/core/parameter/DateWindowingTest.java
delei: Yes,only xlsx is supported,writing to xls is not supported.If the provided files have been set up using MS Excel, there will be no problem reading xls files.
I have provided two xls files that have been set up for testing the reading of xls files.

alaahong (approved):

fastexcel-test/src/test/java/cn/idev/excel/test/core/parameter/DateWindowingTest.java
delei: I tried using `org.apache.poi.hssf.model.InternalWorkbook#createWorkbook()`, but it couldn't set this option in the actual file. I think Apache POI no longer supports this.

https://github.com/apache/poi/blob/bcc7912c8dc284e6f1b345c9295235ecf4fdce27/poi/src/main/java/org/apache/poi/hssf/model/InternalWorkbook.java#L1269-L1274

alaahong: InternalWorkbook#createWorkbook() always set false ~so I thought above case still involving some confused

delei: :disappointed:Me too. I haven't found a way to set this for xls files yet.

alaahong: In my opinion, the test case should be same behavior if always set false by POI. Might be we can reflect to get the value in unit test, just fetch the value but not set to avoid any hack on private field.

Issue #440, opened 2025-07-29
We are already using the flatten-maven-plugin, maybe we can optimize the project version management.


weihubeats:
### Search before asking

- [x] I searched in the [issues](https://github.com/fast-excel/fastexcel/issues) and found nothing similar.

### Motivation

We are already using the flatten-maven-plugin, maybe we can optimize the project version management.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #441, opened 2025-07-29
[ISSUE #440] maybe we can optimize the project version management

weihubeats:
https://github.com/fast-excel/fastexcel/issues/440

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 
     <groupId>cn.idev.excel</groupId>
     <artifactId>fastexcel-parent</artifactId>
-    <version>1.2.0</version>
+    <version>${revision}</version>
 
     <packaging>pom</packaging>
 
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -91,6 +92,13 @@
 
     <dependencyManagement>
         <dependencies>
+
```
delei, line 95: Why is the version number still 0.0.1?

delei:
> #440

It is better to follow the `pull_request_template` template when filling in a PR content.

pom.xml
weihubeats: The project is currently using version 0.0.1. If we use the latest version, many classes are deleted. I have only optimized the dependency management and have not upgraded the dependency versions.

weihubeats:
fix

delei:
Agree,we need to confirm whether it will affect the release workflow. 
cc @psxjoy

weihubeats:
So will this PR affect any release process?

PR #466, opened 2025-08-01
refactor: optimize the StyleUtil code

weihubeats:
https://github.com/fast-excel/fastexcel/issues/465

## What's changed?
1. Fix the incorrectly printed log level
2. Add a generic function to reduce excessive if statements

## Checklist


alaahong (commented):
Seems not mandatory to create a new Class with new Method?
e.g.  inline usage is enough  
```
 Optional.ofNullable(writeCellStyle.getHidden())
        .ifPresent(cellStyle::setHidden); 
```

And meanwhile, even for new Utils, please also provide unit test at same time.

weihubeats:
We can delete the new utility classes and use internal private methods.

alaahong:
> [x] I have added the necessary unit tests and all cases have passed.
Please also fulfill this as you've added several methods~

pom.xml
```diff
@@ -49,6 +49,9 @@
         <logback-classic.version>1.5.6</logback-classic.version>
         <mockserver.version>5.15.0</mockserver.version>
         <jacoco.version>0.8.13</jacoco.version>
+        <junit.version>5.13.4</junit.version>
+        <mockito.version>4.11.0</mockito.version>
```
alaahong: seems above are all existing dependencies....

psxjoy:
All CI fails

fastexcel/src/main/java/cn/idev/excel/util/StyleUtil.java
```diff
@@ -9,16 +9,13 @@
...
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.poi.common.usermodel.HyperlinkType;
 import org.apache.poi.hssf.usermodel.HSSFFont;
 import org.apache.poi.hssf.usermodel.HSSFRichTextString;
-import org.apache.poi.ss.usermodel.CellStyle;
-import org.apache.poi.ss.usermodel.DataFormat;
-import org.apache.poi.ss.usermodel.Font;
-import org.apache.poi.ss.usermodel.RichTextString;
-import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.ss.usermodel.*;
```
psxjoy: Don't use '*'.

weihubeats:
fix ci

fastexcel/pom.xml
```diff
@@ -69,5 +70,41 @@
...
+        </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>-Xmx4g -Xms4g -XX:+HeapDumpOnOutOfMemoryError</argLine>
```
psxjoy: Why use `4g` properties? And we have already imported `maven-surefire-plugin`.

psxjoy (commented):
Something needs to change.

fastexcel/pom.xml
weihubeats: If JVM memory size is not specified, running tests in CI may result in memory shortage and crash.

weihubeats:
What else needs to be done

pom.xml
weihubeats: Currently, the project does not have separately defined JUnit dependency versions, as most rely on dependency management from spring-boot-starter-test. For non-Spring Boot modules, we do not need to strongly couple with Spring Boot's dependency management.

alaahong: Even you left "right" comment, but do you know that the commit bringing multiple versions into one module?

weihubeats: One solution is to use the same JUnit version as the one in spring-boot-starter-test. How does this approach compare to another, more direct method, which is to also reference spring-boot-starter-test for managing test dependencies? Which of these two solutions do people prefer

psxjoy: It is best to use the original dependencies:)

weihubeats: Now, using spring-boot-starter-test for unified management of testing dependencies.

psxjoy (approved):
LGTM
If no one has other questions, this PR will merge in 24h.

Issue #665, opened 2025-10-30
[Bug] An UnsupportedOperationException occurred when setting the dynamic header


delei:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar.

### Fesod version

1.3.0

### JDK version

JDK1.8

### Steps To Reproduce

```java
public class IssueTest {

    @Test
    public void dynamicHeadWrite() throws Exception {
        List<List<String>> multiHeader = new ArrayList<>();
        multiHeader.add(Arrays.asList("stringTitle1"));
        multiHeader.add(Arrays.asList("stringTitle20", "stringTitle21"));

        FastExcel.write(TestFileUtil.createNewFile("dynamicHeadWrite" + System.currentTimeMillis() + ".xlsx"))
                .head(multiHeader)
                .sheet()
                .doWrite(Collections.emptyList());
    }
}

```

### Current Behavior

```bash
java.lang.UnsupportedOperationException
	at java.util.AbstractList.add(AbstractList.java:148)
	at java.util.AbstractList.add(AbstractList.java:108)
	at cn.idev.excel.metadata.property.ExcelHeadProperty.initHeadRowNumber(ExcelHeadProperty.java:108)
	at cn.idev.excel.metadata.property.ExcelHeadProperty.<init>(ExcelHeadProperty.java:88)
	at cn.idev.excel.write.property.ExcelWriteHeadProperty.<init>(ExcelWriteHeadProperty.java:65)
	at cn.idev.excel.write.metadata.holder.AbstractWriteHolder.<init>(AbstractWriteHolder.java:237)
	at cn.idev.excel.write.metadata.holder.WriteWorkbookHolder.<init>(WriteWorkbookHolder.java:186)
	at cn.idev.excel.context.WriteContextImpl.initCurrentWorkbookHolder(WriteContextImpl.java:139)
	at cn.idev.excel.context.WriteContextImpl.<init>(WriteContextImpl.java:117)
	at cn.idev.excel.write.ExcelBuilderImpl.<init>(ExcelBuilderImpl.java:54)
	at cn.idev.excel.ExcelWriter.<init>(ExcelWriter.java:55)
	at cn.idev.excel.write.builder.ExcelWriterBuilder.build(ExcelWriterBuilder.java:151)
	at cn.idev.excel.write.builder.ExcelWriterBuilder.sheet(ExcelWriterBuilder.java:167)
[...]

PR #668, opened 2025-10-31
fix: support immutable lists in dynamic header

bengbengbalabalabeng:

Fixed: #665 

Avoid `UnsupportedOperationException` when using immutable list (such as: `Arrays.asList`、`List.of` or `Collections.unmodifiableList` ...) in dynamic header

## What's changed?

- Modified `AbstractParameterBuilder#head` to always convert `head` into a mutable deep copy, preventing `UnsupportedOperationException` with immutable lists.  
- Added unit tests (`ImmutableListHeadDataTest`).  

## Checklist


fesod/src/test/java/org/apache/fesod/excel/head/ImmutableListHeadDataListener.java
```diff
@@ -0,0 +1,64 @@
...
+        List<List<String>> head = context.readSheetHolder().getHead();
+
+        Assertions.assertInstanceOf(ArrayList.class, head);
+        for (List<String> item : head) {
+            Assertions.assertInstanceOf(ArrayList.class, item);
+        }
+
+        Assertions.assertEquals(1, list.size());
+        Map<Integer, String> data = list.get(0);
+        Assertions.assertEquals("字符串0", data.get(0));
```
delei: It is best to use English.

fesod/src/test/java/org/apache/fesod/excel/head/ImmutableListHeadDataTest.java
```diff
@@ -0,0 +1,95 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fesod.excel.head;
+
+import java.io.File;
+import java.text.ParseException;
+import java.util.*;
```
delei: Avoid using `import *`

bengbengbalabalabeng:
Thanks for the review. I’ve updated the test code as suggested.

fesod/src/main/java/org/apache/fesod/sheet/metadata/AbstractParameterBuilder.java
```diff
@@ -39,10 +41,23 @@ public abstract class AbstractParameterBuilder<T extends AbstractParameterBuilde
...
+     * Ensures and returns a fully mutable deep copy of head list.
+     *
+     * @param head head The source list to create a mutable copy from.
+     * @return A new, fully mutable deep copy, or the original list if the input is null or empty.
+     */
+    private List<List<String>> toMutableListIfNecessary(List<List<String>> head) {
+        if (null == head || head.isEmpty()) {
+            return head;
+        }
+        return head.stream().map(ArrayList::new).collect(Collectors.toList());
```
bengbengbalabalabeng: I’ve replaced the previous stream implementation with a simple for loop to create the copy. I also updated the method description: changed "fully mutable deep copy ..." to "fully mutable of head list" which more accurately reflects the actual behavior.

alaahong: why not directly new ArrayList as no any map or reduce required?

bengbengbalabalabeng:
> Concern on the eager to new, any potential performance impact here?
> 
> And meanwhile, `head.stream().map(ArrayList::new).collect(Collectors.toList())` is not the deep copy.

I considered another approach in #665:

In `cn.idev.excel.metadata.property.ExcelHeadProperty#ExcelHeadProperty` constructor.
- Wrap `head.get(i)` with `ArrayList<String>`. 
- But since the constructor may run multiple times during debugging, this could lead to repeated wrapping.

So I decided not to adopt that solution.

alaahong (commented):
Concern on the eager to new, any potential performance impact here?

And meanwhile, `head.stream().map(ArrayList::new).collect(Collectors.toList())` is not the deep copy.

delei (approved):
LGTM

PR #685, opened 2025-11-12
test(core): improve assertion usage for better clarity and order

GOODBOY008:
### Corrected assertion parameter order
Changed from assertEquals(actual, expected) to assertEquals(expected, actual)
Follows JUnit conventions where expected value comes first
Results in clearer failure messages: "Expected: X, but was: Y"


GOODBOY008:
@alaahong PTAL, Thanks.

psxjoy (approved):
lgtm

PR #688, opened 2025-11-14
refactor: rename and deprecate legacy core classes

delei:
## Purpose of the pull request

Close #651 

## What's changed?

- Modify the inheritance relationship of the alias class
- Remove FesodSheetFactory class
- Modify the comment information of the class

## Checklist

- [x] I have read the [Contributor Guide](https://fesod.apache.org/community/contribution/).
- [x] I have written the necessary doc or comment.
- [x] I have added the necessary unit tests and all cases have passed.

PR #699, opened 2025-11-20
javadoc: Deprecate legacy classes and add migration guidance to FesodSheet

GOODBOY008:
## Description

This PR adds deprecation notices to `EasyExcel`, `FastExcel`, and `FastExcelFactory` classes, providing clear migration guidance to use `FesodSheet`. It also enhances the Javadoc for `FesodSheet` and `FesodSheetFactory`.

## Motivation

As part of the rebranding from FastExcel to Fesod, it is important to guide users away from the old entry points and towards the new `FesodSheet` API. This ensures a smoother transition and better developer experience.

### DIFF
--- a/fesod/src/main/java/org/apache/fesod/sheet/EasyExcel.java
+++ b/fesod/src/main/java/org/apache/fesod/sheet/EasyExcel.java
@@ -20,11 +20,50 @@
 package org.apache.fesod.sheet;
 
 /**
- * An alias class for {@link FesodSheet}.
- * This class is deprecated; use {@link FesodSheet} instead.
- * <p>
- * This class will be removed in future versions.
- * </p>
+ * Legacy alias for {@link FesodSheet}.
+ *
+ * <p><strong>DEPRECATED:</strong> This class has been deprecated as part of the project's transition
+ * to the Apache Fesod branding. Users should migrate to {@link FesodSheet} for all new development.
+ *
+ * <h3>Migration Guide</h3>
+ * <p>Replace all occurrences of {@code EasyExcel} with {@code FesodSheet} in your codebase:
+ *
+ * <table border="1">
+ *   <caption>Migration Examples</caption>
+ *   <tr>
+ *     <th>Old Code (Deprecated)</th>
+ *     <th>New Code (Recommended)</th>
+ *   </tr>
+ *   <tr>
+ *     <td>{@code EasyExcel.read(file, DemoData.class, listener)}</td>
+ *     <td>{@code FesodSheet.read(file, DemoData.class, listener)}</td>
+ *   </tr>
+ *   <tr>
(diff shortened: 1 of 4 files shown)


GOODBOY008:
@psxjoy I think it would be helpful to provide a migration guide to walk users through the migration steps. I’m happy to contribute this migration document to the Fesod website.

delei:
:confused: It appears that such specificity (such as code examples) is not required in the JavaDoc. I recommend adding documentation for this section to the official website when releasing the first ASF version.


BTW, this PR appears to conflict with PR #688 .

GOODBOY008:
@delei Including them can significantly reduce ambiguity and improve developer experience.
Relying solely on the website means users may miss critical context while reading the API. Since JavaDoc is often the first place developers look, providing concrete examples here helps make the API more self-explanatory.
Therefore, I still believe adding these details directly in the JavaDoc is beneficial.

GOODBOY008:
> @psxjoy I think it would be helpful to provide a migration guide to walk users through the migration steps. I’m happy to contribute this migration document to the Fesod website.

@delei Do you have any suggestions on this?

delei:
> @delei Do you have any suggestions on this?

How about adding this to the [Guide](https://fesod.apache.org/docs/quickstart/guide) page?

GOODBOY008:
I suggest adding a separate documentation directory. @delei cc
企业微信截图_25a727e9-f0d7-4cb2-8754-fd5e1ab03b9c

delei:
> I suggest adding a separate documentation directory.

I agree to add it as a separate page, but I recommend a subpage under `Quick Start`, such as the `Version Upgrade Guide`.

GOODBOY008:
> > I suggest adding a separate documentation directory.
> 
> I agree to add it as a separate page, but I recommend a subpage under `Quick Start`, such as the `Version Upgrade Guide`.

@delei Thanks for the suggestion! 

**Quick Start** is for new users, while a **Migration Guide** targets existing users upgrading from an older version. These are different audiences, so nesting migration content inside Quick Start could confuse readers.

For example:
	•	Vue Migration Guide (top-level): https://v2.vuejs.org/v2/guide/migration
	•	JUnit 4 → 5 Migration (top-level): https://docs.junit.org/current/user-guide/#migrating-from-junit4

Keeping Migration Guide as a sibling of Quick Start ensures:
	1.	Clear separation between onboarding and upgrade content
	2.	Easier discoverability for migrating users
	3.	Cleaner, more focused Quick Start section

I recommend placing Migration Guide at the top level rather than inside Quick Start.

@psxjoy Would you like to join this discussion and share your thoughts on this subject?

GOODBOY008:
@psxjoy PTAL

GOODBOY008:
@psxjoy @delei I believe the Javadoc is also important for our developers, as it helps them understand the source code and learn the latest usage guidelines. PTAL~

Issue #703, opened 2025-11-22
[Enhancement] Optimize CI workflows to skip Java tests on docs-only changes


ongdisheng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar.

### Motivation

Currently all Java CI workflows run on every PR even when only documentation files are changed. This wastes CI resources and slows down PR checks. For example, PR #702 only changed `website/community/contribution.md` but still triggered all Java CI tests including license check, spotless, tests across 5 JDK versions, CodeQL scan, etc. 

### Solution

Propose to add paths filtering to `ci.yml` and `codeql-scan.yml` so they only run when Java source code or Maven configs change:
```yml
paths:
  - '.github/workflows/ci.yml'
  - '**/pom.xml'
  - 'fesod/**'
  - 'fesod-*/**'
  - '!**.md'
```

### Anything else?

<img width="778" height="466" alt="Image" src="https://github.com/user-attachments/assets/74872676-59b4-4b63-b1b4-80cf5807788b" />

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

delei:
Hi, @ongdisheng 
Good idea. 

IMO, We could start by making the “Check Licence Header” a standalone workflow. Any PR must require a license header check.

ongdisheng:
> IMO, We could start by making the “Check Licence Header” a standalone workflow. Any PR must require a license header check.

@delei Sure, Iet me work on this.

delei:
Hi, @ongdisheng 
How about providing a separate GitHub workflow for Doc CI? It would only be effective when there are changes to the files in the `website` directory.

FYI, due to the fact that the CI for the preview document still has some issues at present, I suggest that it might be more appropriate to use a single CI for now.

ongdisheng:
Hi @Delei, Thank you for the suggestion! Just to clarify and make sure if I understand correctly: are you suggesting we combine `deploy-docs.yml` and the PR preview workflows into fewer files since they share similar build steps? I'm thinking of creating two consolidated workflows: 
1. `ci-docs.yml`: Combined build and production deployment

<details>
  <summary>Click to expand</summary>

```yaml
name: Documentation CI

on:
  push:
    branches: [main]
    paths: ['website/**']
  pull_request:
    paths: ['website/**']

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
[...]

PR #705, opened 2025-11-22
feat: Skip CI on docs-only changes

ongdisheng:
Related: #703 

This PR optimizes CI workflows to skip unnecessary Java tests when only documentation files are changed.

## What's changed?

This PR added paths filtering to `ci.yml` and `codeql-scan.yml` so they only run when Java code or Maven configs change.

## Checklist

- [ ] I have added the necessary unit tests and all cases have passed.

### DIFF
--- a/DISCLAIMER
+++ b/DISCLAIMER
@@ -1,0 +1,10 @@
+Apache Fesod (incubating) is an effort undergoing incubation at the Apache
+Software Foundation (ASF), sponsored by the Apache Incubator PMC.
+
+Incubation is required of all newly accepted projects until a further review
+indicates that the infrastructure, communications, and decision making process
+have stabilized in a manner consistent with other successful ASF projects.
+
+While incubation status is not necessarily a reflection of the completeness
+or stability of the code, it does indicate that the project has yet to be
+fully endorsed by the ASF.
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,5 @@
+Apache Fesod (Incubating)
+Copyright 2025 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
--- a/lombok.config
+++ b/lombok.config
@@ -1,2 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
 lombok.toString.callSuper = CALL
 lombok.equalsAndHashCode.callSuper= CALL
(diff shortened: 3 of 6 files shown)


GOODBOY008:
Hi @ongdisheng, I think using an ignore-based approach would be cleaner than the include logic.  
Feel free to adopt the implementation from my PR: https://github.com/apache/fesod/pull/711.

ongdisheng:
Hi @GOODBOY008, thank you for working on this optimization! I do agree that the `paths-ignore` approach is simpler and easier to maintain for future changes. However, I noticed it would still trigger CI for several configuration files in our current repository
1. Configuration files: .asf.yaml, .gitignore, licenserc.toml, lombok.config, logo.svg
2. GitHub files: .github/dependabot.yml, .github/ISSUE_TEMPLATE/*.yml
3. Other workflow files: .github/workflows/deploy-docs.yml, markdownlint.yml, etc.
4. Tools: tools/spotless/license-header.txt, mvnw, mvnw.cmd

That's about 20+ files that would still run Java CI builds unnecessarily while the include approach ensures CI only runs for actual code and build changes. I think both approaches have their own trade-offs. Feel free to let me know what do you think?

GOODBOY008:
@ongdisheng The scenarios you listed will still run the CI pipeline and it's very necessary with my implement.Just like action file, if add new one you need to trigger as your implement you will modify the action conditions? It's too complexity.

GOODBOY008:
@ongdisheng  For example, if someone accidentally modifies the mvnw file incorrectly, the CI won’t run and the change could be merged into main, causing CI failures later. Because the PR CI is skipped, this kind of issue cannot be detected in time.

.github/workflows/codeql-scan.yml
```diff
@@ -22,8 +22,20 @@ on:
     branches:
       - master
       - main
+    paths:
+      - '.github/workflows/codeql-scan.yml'
+      - '**/pom.xml'
+      - 'fesod/**'
+      - 'fesod-*/**'
+      - '!**.md'
```
GOODBOY008, line 30:
```suggestion
      - "**.java"
```
Only java file to trigger ci.

.github/workflows/codeql-scan.yml
```diff
@@ -22,8 +22,20 @@ on:
...
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - '.github/workflows/codeql-scan.yml'
+      - '**/pom.xml'
+      - 'fesod/**'
+      - 'fesod-*/**'
+      - '!**.md'
```
GOODBOY008, line 38: Ditto

.github/workflows/codeql-scan.yml
ongdisheng: Hi @GOODBOY008, Thanks for the suggestion! Just wondering, would `"**.java"` actually cover everything we need here? From what I can see, it would only match Java source files, but we also have test resources like Excel and CSV files in `src/test/resources/` that our tests depend on. If someone updates those test files, we probably want CI to run and make sure the tests still pass. Also, `"**.java"` wouldn't catch changes to pom.xml or the CI workflow itself and it would possibly still have the same `mvnw` issue you pointed out earlier. Just wondering if we just add `mvnw` on top of the existing list to cover the Maven wrapper case, would this actually be better?

delei: Thank you all for your discussion. 

IMO, I think this setting should simplify the scene rather than make it more complicated. If in a PR only the resources file or the mvnw file is modified separately, this PR may be directly rejected during the review process.

delei:
I recommend seeking consensus. Both PRs share the same underlying premise, though they approach it differently. We welcome both proposals. 

At present, I believe the `paths-ignore` proposal is the more suitable option.

ongdisheng:
Hi @Delei, Thank you for the feedback. I understand you prefer the paths-ignore approach from #711. I'm happy to close my PR as what matters most is improving the CI efficiency for the project, regardless of whose implementation we use.
@GOODBOY008, thank you for your work on this as well.

PR #707, opened 2025-11-22
docs: adjust markdown quotation marks from tables in the CSV section

delei:
## What's changed?

adjust markdown quotation marks from tables in the CSV section

## Checklist

- [x] I have read the [Contributor Guide](https://fesod.apache.org/community/contribution/).
- [x] I have written the necessary doc or comment.
- [x] I have added the necessary unit tests and all cases have passed.

Issue #678, opened 2025-11-06
[Bug] Preview docs workflow blocked by Apache security policy


ongdisheng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar.

### Fesod version

1.3.0

### JDK version

N/A 

### Operating system

N/A

### Steps To Reproduce

1. Create a PR that changes files in `website/` folder
2. `preview-docs.yml` workflow attempts to start
3. Workflow fails at startup with action restriction error

Failed run: https://github.com/apache/fesod/actions/runs/19072405399

### Current Behavior

The workflow fails at startup because `nwtgck/actions-netlify@v3.0` is blocked by Apache's security policy:

> "The action nwtgck/actions-netlify@v3.0 is not allowed in apache/fesod because all actions must be from a repository owned by your enterprise..."

This prevents preview deployments for documentation PRs.

### Expected Behavior

Preview deployments should work for documentation PRs.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

delei:
We have noted this issue. As we have not yet applied for NETLIFY, please await community approval for this action. Once approved, we shall maintain the Actions secrets configuration within the repository.

ongdisheng:
Hi @delei, thank you so much for looking into this issue! I think the problem might be a bit different from the secrets configuration. The workflow is actually failing at startup before it even gets to check for secrets.

When you look at the [failed run](https://github.com/apache/fesod/actions/runs/19072405399), you'll notice it shows `Startup failure`. This happens because Apache's enterprise [security policy](https://infra.apache.org/github-actions-policy.html) doesn't allow external actions like `nwtgck/actions-netlify@v3.0` to run in `fesod` repository. 

The git submodule approach is a workaround specifically for this security restriction. Instead of referencing the action directly from an external repository which Apache blocks, we can add it as a git submodule in our own repository under `.github/actions/`. This way, GitHub sees it as a local action that lives in the repo, which Apache's policy does allow. I noticed other Apache projects like `apache/apisix-website` use this same submodule [solution](https://github.com/apache/apisix-website/tree/master/.github/actions). 

You're absolutely right that we'll need the Netlify secrets configured once this is approved. I just wanted to mention that currently the workflow can't start due to the action being blocked, but once this fix is in place, we can move forward with the secrets configuration. Hope this helps clarify! Please let me know if you have any questions or if there's anything I should adjust.

delei:
[...]

PR #709, opened 2025-11-23
docs: update example.md to clarify preview instructions

alaahong:

#707 

## Checklist


alaahong:
@ongdisheng  can you help to check the reason why four netlify job trigger here? And any gap on the preview ci?

ongdisheng:
Hi @alaahong, I checked the Netlify logs and found the issue. The deployment is failing with this error: 

> No package.json was found in "/opt/build/repo" 

This happens because Netlify is trying to run `pnpm build` from the repository root. Looking at the Netlify build settings, they appear to have been changed or not configured properly. Since I don't have access to the `fesod` Netlify project to verify or update these settings. Could you help check what the current configuration is? And also since we're already using GitHub Actions to handle the build and deployment through the workflow, the Netlify automatic builds should be disabled to avoid conflicts. This is my current repo project build settings:
image



Regarding your commit with `pull_request_target`, I've been looking into this and found there are some security concerns with that approach. I've researched a two-workflow [pattern](https://github.com/apache/fesod/issues/678#issuecomment-3567559178) and implemented it on my fork. Feel free to check it out and give any feedback.

ongdisheng:
Hi @alaahong, I checked the Netlify logs and noticed something interesting. Your deploy URL shows:
`app.netlify.com/projects/fesod/...`

When I checked @delei recent deployment on PR #707, it shows `app.netlify.com/projects/apache-fesod/....` instead. This would probably mean there are two separate Netlify projects connected to `fesod` repository, which explains why you're seeing multiple Netlify jobs triggering for each PR.

alaahong:
> Hi @alaahong, I checked the Netlify logs and noticed something interesting. Your deploy URL shows: `app.netlify.com/projects/fesod/...`
> 
> When I checked @delei recent deployment on PR #707, it shows `app.netlify.com/projects/apache-fesod/....` instead. This would probably mean there are two separate Netlify projects connected to `fesod` repository, which explains why you're seeing multiple Netlify jobs triggering for each PR.

Can you share the working configuration? 
As delei version still meet issue and I set another project as fesod for this, actually we didn't find out the correct way yet.

alaahong:
> Hi @alaahong, I checked the Netlify logs and found the issue. The deployment is failing with this error:
> 
> > No package.json was found in "/opt/build/repo"
> 
> This happens because Netlify is trying to run `pnpm build` from the repository root. Looking at the Netlify build settings, they appear to have been changed or not configured properly. Since I don't have access to the `fesod` Netlify project to verify or update these settings. Could you help check what the current configuration is? And also since we're already using GitHub Actions to handle the build and deployment through the workflow, the Netlify automatic builds should be disabled to avoid conflicts. This is my current repo project build settings: image
> 
> Regarding your commit with `pull_request_target`, I've been looking into this and found there are some security concerns with that approach. I've researched a two-workflow [pattern](https://github.com/apache/fesod/issues/678#issuecomment-3567559178) and implemented it on my fork. Feel free to check it out and give any feedback.

Both pull_request and pull_request_target won't work as expected...

ongdisheng:
> Both pull_request and pull_request_target won't work as expected...

Hi @alaahong, you're right that `pull_request_target` isn't working yet because it's not on the main branch. Actually, both `pull_request_target` and the `workflow_run` pattern I'm proposing need to be merged to main before they work properly. The key difference is security where `pull_request_target` has security risks because it checks out and runs PR code while secrets are available. The two-workflow pattern with `workflow_run` is safer because it separates the untrusted code execution from the secrets access.

alaahong:
> > Both pull_request and pull_request_target won't work as expected...
> 
> Hi @alaahong, you're right that `pull_request_target` isn't working yet because it's not on the main branch. Actually, both `pull_request_target` and the `workflow_run` pattern I'm proposing need to be merged to main before they work properly. The key difference is security where `pull_request_target` has security risks because it checks out and runs PR code while secrets are available. The two-workflow pattern with `workflow_run` is safer because it separates the untrusted code execution from the secrets access.

As token/site id had been resolved now. Can you try to confirm the right configuration on Netlify and CI in Github Action? 
This PR is used to verify the result, you can raise a new PR if any modify required later.

ongdisheng:
Hi @alaahong, thanks for confirming the token/site ID are resolved! Looking at the Netlify build settings screenshot you shared, I notice it has:

> Build command: pnpm build
> Publish directory: ./website/build

However, this configuration would require setting base directory=`website` and a `pnpm-lock.yaml` file in the website folder, which we don't have. This is likely why the builds are failing. 
image


I have a suggestion: Since either `pull_request_target` or `workflow_run` approach need to be merged to main before they can work anyway, how about I submit a PR with the secure `workflow_run` approach I've tested on my fork? This approach uses GitHub Actions to handle the entire build and deployment process, so we can leave the Netlify build settings empty similar to what I showed earlier. Once it's merged to main, we can test it with your PR to verify everything works correctly. Let me know what do you think about this?

ongdisheng:
Hi @alaahong @delei, below is the current configuration I used in my forked repo: 

1. Netlify Settings: I noticed that the Netlify bot was still commenting on PRs even when there were no website changes. To fix this, I switched off the automatic deploy previews for PR in the Netlify settings.
image
image

2. GitHub Actions Workflows: Since we switched off Netlify's automatic deploy previews, all preview builds are now handled by GitHub Actions. I've implemented a secure two-workflow pattern that only builds when `website/**` files change:
   - [Build Documentation](https://github.com/ongdisheng/fesod/blob/main/.github/workflows/pr-docs-build.yml): Builds the website when `website/**` files change
   - [Preview Documentation](https://github.com/ongdisheng/fesod/blob/main/.github/workflows/pr-docs-preview.yml): Deploys to Netlify and posts a preview comment

## Summary
- The workflows only trigger for PR when `website/**` files change.
- The solution is more secure for fork PRs using the two-workflow pattern instead of using `pull_request_target`.
- Only one GitHub Actions bot comment will appear per PR with the deploy preview URL, which updates automatically on each push.
- All actions use Apache-approved versions with SHA pinning.

## Testing Results on My Own Repo:
- [PR with website changes](https://github.com/ongdisheng/fesod/pull/3): Build triggered and preview deployed with comment
- [PR without website changes](https://github.com/ongdisheng/fesod/pull/2): No build triggered and no comment

Feel free to let me know if you have any questions or suggestions for improvements. Thanks!

PR #780, opened 2026-01-07
feature: Add the support module to meet ASF policy requirements

psxjoy:
Add the `fesod-support` module to meet ASF policy requirements.

## Checklist

### DIFF
--- a/fesod-shaded/pom.xml
+++ b/fesod-shaded/pom.xml
@@ -48,15 +48,6 @@
 
 
     <build>
-        <resources>
-            <resource>
-                <directory>src/main/resources</directory>
-                <excludes>
-                    <exclude>META-INF/LICENSE_BINARY</exclude>
-                    <exclude>META-INF/NOTICE_BINARY</exclude>
-                </excludes>
-            </resource>
-        </resources>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>


dist/licenses/LICENSE-spring-core.txt
```diff
@@ -0,0 +1,202 @@
...
+                        https://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
```
pjfanning, line 12: * let me look into this - there is no guarantee that spring-core is correct
* these classes that we are taking from spring-core come from cglib and asm and we at least should also have their licenses too.
* since this license is just a generic Apache license, I would prefer to put a single LICENSE and NOTICE into META-INF that is basically the Fesod license and notice with additional text saying we also have Apache licensed classes from spring-core - so no spring license, just a modified Fesod license/notice that mentions spring-core and ideally the affected shaded package names.

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/BeanMapUtils.java
```diff
@@ -19,8 +19,8 @@
 
 package org.apache.fesod.sheet.util;
 
-import cn.idev.excel.support.cglib.beans.BeanMap;
-import cn.idev.excel.support.cglib.core.DefaultNamingPolicy;
+import org.apache.fesod.support.cglib.beans.BeanMap;
```
pjfanning: can we make this `org.apache.fesod.shaded.cglib`? - likewise with asm

delei: > can we make this `org.apache.fesod.shaded.cglib`? - likewise with asm

+1

psxjoy:
- rename `fesod-support` to  `fesod-shaded`
- add `spring-framework` license info into `LICENSE` and `NOTICE`
- add `cglib` license txt into `dist`
PTAL @pjfanning  @delei   :)

LICENSE
pjfanning, line 1: I think we need a separate license and notice file that we add to the jar. The main license and notice for our source code  does not need this.

NOTICE
```diff
@@ -14,14 +14,4 @@ The original idea and initial implementation of the maven-wrapper module is deri
...
-
-This product is licensed to you under the Apache License, Version 2.0
-(the "License"). You may not use this product except in compliance with
-the License.
-
-This product may include a number of subcomponents with separate
-copyright notices and license terms. Your use of the source code for
-these subcomponents is subject to the terms and conditions of the
-subcomponent's license, as noted in the license.txt file.
+This product includes software developed by the Spring Framework Project (https://spring.io).
```
pjfanning: Same as license comment. We should have a separate notice file for the jar.

psxjoy: > Same as license comment. We should have a separate notice file for the jar.

Done. PTAL

fesod-shaded/src/main/resources/META-INF/LICENSE
pjfanning, line 1: File can be called LICENSE. No need to suffix it with 'binary'.

fesod-shaded/src/main/resources/META-INF/NOTICE_BINARY
```diff
@@ -0,0 +1,7 @@
+Apache Fesod (Incubating)
+Copyright 2025 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+-----------------------------------------------------------------------
+This product includes software developed by the Spring Framework Project (https://spring.io).
```
pjfanning: Add new line

fesod-shaded/src/main/resources/META-INF/LICENSE
```diff
@@ -0,0 +1,218 @@
...
+
+   The Apache Fesod(incubating) project contains subcomponents with separate copyright
+   notices and license terms. Your use of the source code for these
+   subcomponents is subject to the terms and conditions of the following
+   licenses.
+
+   ========================================================================
+    Apache License 2.0
+   ========================================================================
+   This product bundles modified portions of the Spring Framework,
```
pjfanning, line 214: the asm code was originally BSD licensed - I would suggest adding the full text of license-asm.txt in here too

psxjoy: > the asm code was originally BSD licensed - I would suggest adding the full text of license-asm.txt in here too

Thanks for pointing that out. I have updated the LICENSE file to include the full text of the ASM BSD license as suggested.

fesod-shaded/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+            </exclusions>
+        </dependency>
+    </dependencies>
+
+
+    <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <excludes>
```
pjfanning: what is this exclude for?

psxjoy: > what is this exclude for?

I sincerely apologize for not deleting the incorrect file name earlier. I have resubmitted the PR.

PR #901, opened 2026-04-22
[Test] Add testkit infrastructure and refactor unit tests with JUnit 5 parameterized patterns

GOODBOY008:
## Summary

Implements Phase 1-6 of the test modernization proposal from issue #717.

### New Testkit Infrastructure (`testkit/` package)
- `AbstractExcelTest` — base class with format providers and temp file management
- `ExcelFormat` enum — XLSX/XLS/CSV with capability metadata
- `CollectingReadListener` — reusable data collector without embedded assertions
- `ExcelAssertions` — fluent Excel file assertions via AssertJ

### Parameterized Tests (JUnit 5)
- Replace 3 nearly identical per-format test methods with a single `@ParameterizedTest`
- Automatic format coverage: XLSX, XLS, and CSV in one method

### Separation of Concerns
- Remove listener classes that embed `Assertions.*` calls (SRP violation)
- Move all assertions into test methods

### Package Reorganization
- Consolidate legacy scattered packages into `core/`, `format/`, `readwrite/`, `sheet/`, `style/`
- Merge near-duplicate model classes

## Metrics

| | Before | After |
|---|---|---|
| Listener Classes | 25+ | 1 |
| Per-format test methods | 3x | 1x (parameterized) |

## Related

Closes #717

PR #906, opened 2026-04-30
test: cover field-level converter precedence

skytin1004:

I've added test coverage for converter precedence when both a field-level converter and a registered converter could apply to the same Java type.

The behavior I wanted to lock down is that a converter declared directly on a field with `@ExcelProperty(converter = ...)` should take precedence over a converter registered through `registerConverter(...)`.

## What's changed?

Added a regression test in `CustomConverterTest` for this case.

The test writes a CSV with two `String` fields. One field has a field-level converter declared with `@ExcelProperty`, and the other field has no field-level converter, so it should use the registered `String` converter.

The expected output contains both converted values: `field:value,registered:value`.

This verifies the registered converter is actually active for the unannotated field and the field-level converter still takes precedence for the annotated field.

This is useful because users may register a general converter as a default behavior, while still expecting field-specific converters to override it for special cases.

Tested with `fesod-sheet -Dmaven.test.skip=false -Dtest=CustomConverterTest test`.

Result: `Tests run: 7, Failures: 0, Errors: 0, Skipped: 0`.

## Checklist

### DIFF
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/CustomConverterTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/CustomConverterTest.java
@@ -20,14 +20,21 @@
 package org.apache.fesod.sheet.converter;
 
 import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


skytin1004:
Just noticed that #901 also changes `CustomConverterTest.java`. If #901 is merged first, I will rebase this PR and move this test to the new structure.

psxjoy (approved):
lgtm

PR #907, opened 2026-04-30
fix: correct CSV physical cell count

skytin1004:

While looking at the CSV row API from a workbook write handler, I noticed that `CsvRow#getPhysicalNumberOfCells()` was returning the row index instead of the number of cells in the row.

I checked the implementation and found that `getPhysicalNumberOfCells()` was calling `getRowNum()`. That means the first row returns `0`, even when the row actually contains cells. For example, a CSV header row with `No`, `Name`, and `Age` should report `3` physical cells, but it reported `0`.

## What's changed?

I changed `CsvRow#getPhysicalNumberOfCells()` to return `cellList.size()` instead of `getRowNum()`.

I also added a regression test in `CsvFormatTest`. The test writes a CSV file with three header cells, then checks the first row from a workbook write handler and verifies that `getPhysicalNumberOfCells()` returns `3`.

Tested with `CsvFormatTest#testPhysicalNumberOfCells`, full `CsvFormatTest`, and `spotless:check`.

## Checklist


delei (approved):
LGTM

PR #911, opened 2026-05-07
fix: avoid recursive hashCode in write holder contexts

skytin1004:

Related: #889

This PR addresses only the recursive `hashCode()` failure path discussed in #889 and #899.

While rechecking the original failure, I found that the custom converter reproduction appears to involve at least two separate concerns:

1. recursive `hashCode()` evaluation in write holder/context objects
2. global write converter lookup when a registered converter declares a non-null `supportExcelTypeKey()`

To keep the review focused, this PR deliberately handles only the first issue.

I reproduced the recursive `hashCode()` paths locally and added a focused regression test for them:
The regression test creates the recursive holder/context references directly and then calls `hashCode()`, so it covers this PR's failure path without depending on converter behavior.

- `WriteWorkbookHolder.hashCode()` -> `WorkbookWriteHandlerContext.hashCode()` -> `WriteWorkbookHolder.hashCode()`
- `WriteWorkbookHolder.hashCode()` -> initialized sheet holder map -> `WriteSheetHolder.hashCode()` -> parent workbook holder -> `WriteWorkbookHolder.hashCode()`

The fix excludes only the recursive runtime references from Lombok-generated `equals/hashCode`:

- `WriteWorkbookHolder.workbookWriteHandlerContext`
- `WriteSheetHolder.parentWriteWorkbookHolder`
- `WriteSheetHolder.hasBeenInitializedTable`
- `WriteTableHolder.parentWriteSheetHolder`

This PR intentionally does not change converter lookup behavior. I plan to handle that separately with a smaller follow-up test/fix, because it can fail independently when a global converter is registered with a non-null `supportExcelTypeKey()`.

I verified the change locally by running:

- the new `WriteHolderHashCodeTest`
- the existing `CustomConverterTest`
- the full `fesod-sheet` test suite

All tests passed.

> [!NOTE]
>
> If this recursive `hashCode()` fix looks good and gets merged, I can follow up with a smaller PR focused only on the global write converter lookup case, so that the remaining part of #889  can be addressed clearly.

## Checklist


fesod-sheet/src/test/java/org/apache/fesod/sheet/write/metadata/holder/WriteHolderHashCodeTest.java
```diff
@@ -0,0 +1,66 @@
...
+        WriteSheetHolder sheetHolder = new WriteSheetHolder(writeSheet(), workbookHolder);
+        WriteTableHolder tableHolder = new WriteTableHolder(writeTable(), sheetHolder);
+
+        workbookHolder.getHasBeenInitializedSheetIndexMap().put(sheetHolder.getSheetNo(), sheetHolder);
+        workbookHolder.getHasBeenInitializedSheetNameMap().put(sheetHolder.getSheetName(), sheetHolder);
+        sheetHolder.getHasBeenInitializedTable().put(tableHolder.getTableNo(), tableHolder);
+
+        Assertions.assertDoesNotThrow(workbookHolder::hashCode);
+        Assertions.assertDoesNotThrow(sheetHolder::hashCode);
+        Assertions.assertDoesNotThrow(tableHolder::hashCode);
```
alaahong, line 51: any change on the equals behavior？

skytin1004: @alaahong Yes, this also changes equals behavior. That was intentional, because Lombok applies the exclude list to both equals() and hashCode().

The excluded fields are only used during the write process and they can form cycles. So if we only think about hashCode(), the same kind of recursion risk can still remain in equals().

If you prefer, I can update the PR description and add an equals regression assertion to make this explicit.

alaahong: @skytin1004  please kindly help if would like to update this. it's optional for this PR.

delei (approved):
LGTM

PR #912, opened 2026-05-07
fix: support SQL date types in WriteCellData

### LINKED ISSUE #714: [Bug] WriteCellData 方法如果传入 java.sql.Date 会报错
### Search before asking

### Fesod version

1.3.0

### JDK version

17

### Operating system

_No response_

### Steps To Reproduce

Mybatis 3.5.17 之后的版本，date类型的数据返回的是 java.sql.Date 而不是 java.util.Date，此时在 WriteCellData 方法中调用 dateValue.toInstant() 会报错

```java
    /**
     * Constructor for creating a WriteCellData object with a Date value.
     *
     * @param dateValue The Date value to be written to the cell.
     * @throws IllegalArgumentException If the date value is null.
     */
    public WriteCellData(Date dateValue) {
        super();
        if (dateValue == null) {
            throw new IllegalArgumentException("DateValue can not be null");
        }
        setType(CellDataTypeEnum.DATE);
        this.dateValue = LocalDateTime.ofInstant(dateValue.toInstant(), ZoneId.systemDefault());
    }
```

参考链接 
https://github.com/mybatis/mybatis-3/commit/28af7e9d72a9e4952b6ec2f76d4ce12e030cb20a#diff-3100284b93f6e47f7e09229bac897dc14edbb2e162002f24e804113f49d8b6a6R44-R52

Image

Image

### Current Behavior

throw new java.lang.UnsupportedOperationException();

### Expected Behavior

正常

### Anything else?

_No response_

skytin1004:

Closed: #714

I've fixed `WriteCellData(Date)` when the actual value is `java.sql.Date` or `java.sql.Time`.

I reproduced the failure locally and confirmed that both SQL date/time types can throw `UnsupportedOperationException` when `toInstant()` is called directly.

## What's changed?

The existing `toInstant()` path is kept for `java.util.Date` and `java.sql.Timestamp`. This keeps the behavior unchanged for regular date values and preserves `java.sql.Timestamp` nano precision.

I added regression tests for:

- `java.sql.Date`
- `java.sql.Time`
- `java.sql.Timestamp`

To verify the fix, I first confirmed that the new `WriteCellDataTest` failed before the change with `UnsupportedOperationException` from `java.sql.Date.toInstant()` and `java.sql.Time.toInstant()`. After applying the fix, I confirmed that the new test passed, the existing related `CellDataDataTest` passed together with it, and the full `fesod-sheet` test suite also passed.

## Checklist


fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/data/WriteCellData.java
```diff
@@ -169,7 +170,11 @@ public WriteCellData(Date dateValue) {
             throw new IllegalArgumentException("DateValue can not be null");
         }
         setType(CellDataTypeEnum.DATE);
-        this.dateValue = LocalDateTime.ofInstant(dateValue.toInstant(), ZoneId.systemDefault());
+        if (dateValue instanceof java.sql.Date || dateValue instanceof java.sql.Time) {
+            this.dateValue = LocalDateTime.ofInstant(Instant.ofEpochMilli(dateValue.getTime()), ZoneId.systemDefault());
+        } else {
+            this.dateValue = LocalDateTime.ofInstant(dateValue.toInstant(), ZoneId.systemDefault());
+        }
```
alaahong, line 178: is any time zone issue here? seems depends on the execution machine

skytin1004: @alaahong Good point. I changed this path to use SQL Date/Time APIs directly instead of going through epoch millis.

java.sql.Date now uses toLocalDate(), and java.sql.Time uses toLocalTime(). The existing toInstant() path is still kept for java.util.Date and java.sql.Timestamp.

I also checked the other date conversions. Some existing generic Date/Calendar paths still use the system default timezone, but they are not introduced by this PR. I will look at those separately and follow up if we want to make them consistent.

delei (approved):
LGTM

PR #936, opened 2026-06-16
feat: add FesodDoc classes with initial structure

psxjoy:
## Checklist


bengbengbalabalabeng (commented):
We could try to move the `<licenseHeader>` configuration of `spotless-maven-plugin` down to the current module, so it only applies to the `fesod-doc` files, avoiding any impact on historical files in other existing modules while also eliminating the need to manually maintain license headers.

PR #954, opened 2026-07-17
fix: CsvCell crashes on java.sql.Date/Time due to unsupported toInstant()

alaahong:

The SQL date type fix in #912  was only applied to WriteCellData, not
to CsvCell.setCellValueImpl(Date). The old code calls value.toInstant()
unconditionally, which throws UnsupportedOperationException for
java.sql.Date and java.sql.Time on Java 9+ (these types override
toInstant() to throw because they represent date-only/time-only values).

Fix: Add the same instanceof checks as WriteCellData - use toLocalDate()
.atStartOfDay() for java.sql.Date and toLocalTime().atDate(...) for
java.sql.Time, falling back to toInstant() for regular java.util.Date.

Before Change
image

## What's changed?

CsvCell

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/csv/CsvCell.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/csv/CsvCell.java
@@ -164,7 +164,13 @@ public class CsvCell extends CellBase {
         if (value == null) {
             return;
         }
-        this.dateValue = LocalDateTime.ofInstant(value.toInstant(), ZoneId.systemDefault());
+        if (value instanceof java.sql.Date) {
+            this.dateValue = ((java.sql.Date) value).toLocalDate().atStartOfDay();
+        } else if (value instanceof java.sql.Time) {
+            this.dateValue = ((java.sql.Time) value).toLocalTime().atDate(java.time.LocalDate.of(1970, 1, 1));
+        } else {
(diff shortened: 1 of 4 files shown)


fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/csv/CsvCell.java
```diff
@@ -164,7 +164,13 @@ protected void setCellValueImpl(Date value) {
         if (value == null) {
             return;
         }
-        this.dateValue = LocalDateTime.ofInstant(value.toInstant(), ZoneId.systemDefault());
+        if (value instanceof java.sql.Date) {
+            this.dateValue = ((java.sql.Date) value).toLocalDate().atStartOfDay();
+        } else if (value instanceof java.sql.Time) {
+            this.dateValue = ((java.sql.Time) value).toLocalTime().atDate(java.time.LocalDate.of(1970, 1, 1));
```
bengbengbalabalabeng: `LocalDate.of(1970, 1, 1)` is recommended to be extracted as a public constant, such as `DateUtils.EPOCH` (due to project environment constraints, the JDK 9+ `LocalDate.EPOCH` cannot be used directly). Using a custom constant can improve semantic clarity while avoiding repeated object creation.
Similarly: #954

bengbengbalabalabeng (approved):
LGTM

PR #957, opened 2026-07-18
fix: clone DataFormatData in WriteCellStyle.merge to avoid mutating cached annotation style

### LINKED ISSUE #956: [Bug] Cell style strategy corrupts the cached @ContentStyle(dataFormat) of other rows
### Search before asking

### Fesod version

current main (4b426ba)

### JDK version

### Operating system

Linux (not OS-specific)

### Steps To Reproduce

A `@ContentStyle(dataFormat = N)` annotation's number format is silently overwritten by a registered cell style strategy, producing the wrong number format in the output file.

```java
@Data
public class DemoData {
    @ExcelProperty("value")
    @ContentStyle(dataFormat = 4) // "#,##0.00"
    private Double value;
}

// content style list alternating between a style WITH a data format and one WITHOUT
WriteCellStyle withFormat = new WriteCellStyle();
DataFormatData dfd = new DataFormatData();
dfd.setIndex((short) 2); // "0.00"
withFormat.setDataFormatData(dfd);
WriteCellStyle withoutFormat = new WriteCellStyle();

FesodSheet.write(file, DemoData.class)
        .registerWriteHandler(new HorizontalCellStyleStrategy(null, Arrays.asList(withFormat, withoutFormat)))
        .sheet()
        .doWrite(Arrays.asList(new DemoData(1111.5), new DemoData(2222.5), new DemoData(3333.5)));
```

The second data row's strategy style defines **no** data format, so that cell should fall back to the annotation's `#,##0.00`.

### Current Behavior

All three data rows come out with format index 2 (`0.00`). Reading the written file back with POI:

```
dataRow0 format = 2 (expected 2)
dataRow1 format = 2 (expected 4 - annotation)   <-- wrong
dataRow2 format = 2 (expected 2)
```

With `.filedCacheLocation(CacheLocationEnum.MEMORY)` it is worse: the corruption leaks into **later, completely unrelated writes** of the same bean class - a follow-up write with *no* custom handlers at all also emits format 2 instead of the annotation's 4, because the static `ClassUtils.CONTENT_CACHE` now holds the mutated format.

### Expected Behavior

Rows whose strategy style does not define a data format keep the `@ContentStyle(dataFormat = 4)` annotation format (`#,##0.00`), and the cached annotation metadata is never mutated by a style strategy.

### Anything else?

Root cause: `WriteCellStyle.merge` aliases the source's `DataFormatData` into the target without cloning:

https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/style/WriteCellStyle.java#L180-L186

```java
if (target.getDataFormatData() == null) {
    target.setDataFormatData(source.getDataFormatData()); // shared reference, no copy
}
```

The annotation's `DataFormatData` is parsed once and cached. After this merge, the per-cell style points at that cached object, so when a second style source merges into the same cell, `DataFormatData.merge` writes into it and permanently changes the cached annotation format.

The `writeFont` field a few lines below already handles this correctly by cloning:

```java
target.setWriteFont(source.getWriteFont().clone());
```

Doing the same for `DataFormatData` fixes it:

```java
target.setDataFormatData(source.getDataFormatData().clone());
```

nkuprins:

Closed: #956

## What's changed?

`WriteCellStyle.merge` copied the source's `DataFormatData` into the target **by reference**. Because the annotation-driven style source is parsed once and cached, a later merge from another style source (e.g. a registered `HorizontalCellStyleStrategy`) wrote into that shared object and permanently changed the cached `@ContentStyle(dataFormat)` - producing the wrong number format on rows where the strategy defines no format, and (with `CacheLocationEnum.MEMORY`) on later, unrelated writes of the same bean class.

The fix clones the `DataFormatData` before assigning it, exactly like `writeFont` is already cloned a few lines below in the same method:

```diff
- target.setDataFormatData(source.getDataFormatData());
+ target.setDataFormatData(source.getDataFormatData().clone());
```

Added `ContentStyleDataFormatPollutionTest` with two regression tests that assert on the number format of the **written file** (via the testkit's `ExcelAssertions`):

- `annotationFormatSurvivesStyleStrategyInSameWrite` - default config: a row whose strategy style has no data format must keep the annotation's `#,##0.00`.
- `annotationFormatSurvivesAcrossWritesWithMemoryCache` - `MEMORY` cache location: a follow-up write with no custom handlers must keep the annotation's format.

Both tests fail on `main` (format comes out `0.00`) and pass with the fix. The full `fesod-sheet` suite passes with the fix applied (673 tests, 0 failures, 0 errors, 0 skipped), confirming no behavior change for styles that don't involve a shared `DataFormatData`.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/style/WriteCellStyle.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/style/WriteCellStyle.java
@@ -179,7 +179,7 @@ public class WriteCellStyle {
         }
         if (source.getDataFormatData() != null) {
             if (target.getDataFormatData() == null) {
-                target.setDataFormatData(source.getDataFormatData());
+                target.setDataFormatData(source.getDataFormatData().clone());
             } else {
                 DataFormatData.merge(source.getDataFormatData(), target.getDataFormatData());
             }
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/style/ContentStyleDataFormatPollutionTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/style/ContentStyleDataFormatPollutionTest.java
@@ -30,11 +30,13 @@ import org.apache.fesod.sheet.annotation.ExcelProperty;
 import org.apache.fesod.sheet.annotation.write.style.ContentStyle;
 import org.apache.fesod.sheet.enums.CacheLocationEnum;
 import org.apache.fesod.sheet.metadata.data.DataFormatData;
+import org.apache.fesod.sheet.testkit.Tags;
 import org.apache.fesod.sheet.testkit.assertions.ExcelAssertions;
 import org.apache.fesod.sheet.testkit.base.AbstractExcelTest;
 import org.apache.fesod.sheet.testkit.enums.ExcelFormat;
 import org.apache.fesod.sheet.write.metadata.style.WriteCellStyle;
 import org.apache.fesod.sheet.write.style.HorizontalCellStyleStrategy;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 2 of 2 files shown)


bengbengbalabalabeng (approved):
LGTM.

merge in 24h

PR #964, opened 2026-07-20
refactor: encapsulate xlsx boolean cell parsing in CellData

### LINKED ISSUE #953: [Enhancement] Pin intended behavior of BooleanUtils.valueOf("true") with a test?
### Search before asking

### Motivation

`BooleanUtils.valueOf(String)` in `fesod-common` returns `TRUE` only for the exact string `"1"` - every other input, including `"true"`, maps to `FALSE`:

```java
public static Boolean valueOf(String str) {
    if (TRUE_NUMBER.equals(str)) {  // TRUE_NUMBER = "1"
        return Boolean.TRUE;
    }
    return Boolean.FALSE;
}
```

Its only production caller is `CellTagHandler`, converting the `<v>` value of `t="b"` (boolean) cells when streaming xlsx. The strict `"1"` only check matches Excel's own output (Excel always writes `0`/`1`) and Apache POI's DOM model (`XSSFCell.getBooleanCellValue()` does the same `"1".equals(v)`), so it looks intentional.

However, `BooleanUtilsTest` only covers `"1"`, `""`, and `null`. Nothing in the repo documents that `valueOf("true") == FALSE` is by design. Notably, POI's own SAX path (`XSSFSheetXMLHandler`) is more lenient (`value.charAt(0) == '0' ? FALSE : TRUE`), so the precedent is mixed and a future contributor could plausibly "fix" this as a bug and change reading behavior.

**Question for the maintainers:** is the `"1"` only matching the intended contract? If yes, would you accept a small test-only PR pinning it?

### Solution

If the current behavior is intended, add assertions to `BooleanUtilsTest` documenting it, e.g.:

```java
// Intentional: xlsx boolean cells use Excel's 0/1 markers; matches POI XSSFCell
Assertions.assertFalse(BooleanUtils.valueOf("true"));
Assertions.assertFalse(BooleanUtils.valueOf("0"));
```

optionally plus a one-line Javadoc note on `valueOf` stating it targets Excel's `0`/`1` boolean cell markers.

### Alternatives

If instead the maintainers consider the strictness a defect (files written by non-Excel producers could in principle contain `<v>true</v>`), the alternative is making valueOf lenient - e.g. also accepting "true" case-insensitively, similar to POI's XSSFSheetXMLHandler. That would be a behavior change to xlsx boolean reading, so I did not assume it.

### Anything else?

_No response_

nkuprins:

Closed: #953

Follows up on the discussion in #953. As suggested by @bengbengbalabalabeng, this encapsulates the xlsx boolean-cell parsing where it's actually used instead of exposing it as a shared utility.

## What's changed?

BooleanUtils.valueOf(String) returned true only for the exact string "1"; every other input (including "true") mapped to false. That is correct for its one job - parsing the <v> text of t="b" boolean cells, where Excel writes 0/1, matching POI's XSSFCell.getBooleanCellValue() - but as a public util named valueOf it reads like a bug and invites a future contributor to "fix" it into leniency, silently changing xlsx read behavior.

- Add CellData.setBooleanValueFromString(String) holding the "1" ->  true logic (private TRUE_NUMBER constant).
- Rewire the sole caller, CellTagHandler (xlsx SAX read path), to use it. Drop the now-unused BooleanUtils import.
- Deprecate BooleanUtils.valueOf(String) - not a general-purpose string-to-boolean conversion, no other callers - pointing to the new method for future removal.
- Add CellDataTest pinning the behavior.

No behavior change: the parsing logic is identical, just relocated onto CellData. The deprecated valueOf still works for any external caller until it's removed.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/data/CellData.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/data/CellData.java
@@ -45,6 +45,11 @@ import org.apache.fesod.sheet.metadata.AbstractCell;
 @EqualsAndHashCode
 public class CellData<T> extends AbstractCell {
     /**
+     * Marker Excel writes for a {@code true} boolean cell value ({@code <v>1</v>} for {@code t="b"} cells).
+     */
+    private static final String TRUE_NUMBER = "1";
+
+    /**
      * cell type
      */
     private CellDataTypeEnum type;
@@ -70,6 +75,19 @@ public class CellData<T> extends AbstractCell {
      * formula
      */
     private FormulaData formulaData;
+
+    /**
+     * Sets {@link #booleanValue} from the raw {@code <v>} text of a boolean ({@code t="b"}) xlsx cell.
+     *
+     * <p>Excel encodes boolean cells using the numeric markers {@code 0}/{@code 1}, so only the exact
+     * string {@code "1"} maps to {@code true}; any other input (including {@code "true"}) maps to
+     * {@code false}. This mirrors Apache POI's {@code XSSFCell.getBooleanCellValue()}.
+     *
+     * @param str the raw cell value text
+     */
+    public void setBooleanValueFromString(String str) {
+        this.booleanValue = TRUE_NUMBER.equals(str);
+    }
 
     /**
      * Ensure that the object does not appear null
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/data/CellData.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/data/CellData.java
@@ -45,11 +45,6 @@ import org.apache.fesod.sheet.metadata.AbstractCell;
 @EqualsAndHashCode
 public class CellData<T> extends AbstractCell {
     /**
-     * Marker Excel writes for a {@code true} boolean cell value ({@code <v>1</v>} for {@code t="b"} cells).
-     */
-    private static final String TRUE_NUMBER = "1";
-
-    /**
      * cell type
      */
     private CellDataTypeEnum type;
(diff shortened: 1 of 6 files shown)


bengbengbalabalabeng:
There are some conflicts in the current branch. Please resolve them.

nkuprins:
> There are some conflicts in the current branch. Please resolve them.

Done. I can also add you as a _co-author_ to the commit if you'd like, given your suggestions on the implementation :)

bengbengbalabalabeng:
> Done. I can also add you as a _co-author_ to the commit if you'd like, given your suggestions on the implementation :)

Thanks for the co-author offer. I appreciate it, but the implementation is your work, so no need to add me.

bengbengbalabalabeng (dismissed):
LGTM

Merge in 24h.

fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/data/CellData.java
```diff
@@ -71,6 +76,19 @@ public class CellData<T> extends AbstractCell {
...
+    /**
+     * Sets {@link #booleanValue} from the raw {@code <v>} text of a boolean ({@code t="b"}) xlsx cell.
+     *
+     * <p>Excel encodes boolean cells using the numeric markers {@code 0}/{@code 1}, so only the exact
+     * string {@code "1"} maps to {@code true}; any other input (including {@code "true"}) maps to
+     * {@code false}. This mirrors Apache POI's {@code XSSFCell.getBooleanCellValue()}.
+     *
+     * @param str the raw cell value text
+     */
+    public void setBooleanValueFromString(String str) {
```
delei: There is no logical issue. 

FYI, I will just add a point from the design perspective. I have noticed that the CellData class is a highly abstract entity. Is it more appropriate for the assignment of internal attributes to be handled in the subclasses `ReadCellData` and `WriteCellData` ?

nkuprins: Good point, agreed. Should I then move `setBooleanValueFromString`  (and its `TRUE_NUMBER` constant) down to `ReadCellData` and commit-push?

delei: SGTM. You can continue to commit or update this PR.

nkuprins: Done

delei (approved):
LGTM

PR #965, opened 2026-07-21
fix: descriptive error for unrecognized cell type attribute

### LINKED ISSUE #955: [Enhancement] Descriptive error for unrecognized cell type attribute instead of "IllegalArgumentException: Type can not be null"
### Search before asking

### Motivation

This is about failing with a useful error, not about supporting more cell types.

A cell whose `t` attribute is not one of `s`, `str`, `inlineStr`, `e`, `b`, `n` aborts the whole read with an error that names neither the cell nor the attribute:
```
java.lang.IllegalArgumentException: Type can not be null
    at org.apache.fesod.sheet.metadata.data.ReadCellData.<init>(ReadCellData.java:78)
    at org.apache.fesod.sheet.analysis.v07.handlers.CellTagHandler.startElement(CellTagHandler.java:63)
```

**Repro:** 
Any xlsx sheet that contains e.g. `<c r="A1" t="d"><v>2024-05-01T00:00:00</v></c>` (or `t="anything"`)

**Cause:** 
In `CellDataTypeEnum.buildFromCellType`, a missing `t` attribute is handled - it returns `EMPTY`. But when `t` is present with a value that isn't in the map, the lookup returns `null` and nothing checks for it. That null then trips the `ReadCellData` constructor, which is where the confusing exception comes from.

### Solution

Keep failing, but say what happened. In `CellTagHandler.startElement`:

```java
String cellType = attributes.getValue(ExcelXmlConstants.ATTRIBUTE_T);
CellDataTypeEnum type = CellDataTypeEnum.buildFromCellType(cellType);
if (type == null) {
    throw new ExcelAnalysisException("Unsupported cell type '" + cellType + "'");
}
```

### Alternatives

Fall back to `DIRECT_STRING` when the type is `null`

### Anything else?

_No response_

nkuprins:

Closed: #955

## What's changed?

An `.xlsx` cell whose `t` attribute isn't a recognized type (`s`, `str`, `inlineStr`, `e`, `b`, `n`) made `CellDataTypeEnum.buildFromCellType` return `null`, which `CellTagHandler` passed straight into `new ReadCellData<>(type)`, leading to `IllegalArgumentException: Type can not be null`, which names neither the cell nor the attribute.

`CellTagHandler.startElement` now detects the `null` and throws an `ExcelAnalysisException` that names the type and the cell:

```java
Invalid cell data type: 'unknown' in cell B4
```

Following the discussion with @bengbengbalabalabeng on the issue, the check lives in the handler (rather than centralised in `CellDataTypeEnum.buildFromCellType`), so the message can point at the exact cell. And `CellDataTypeEnum.buildFromCellType` Javadoc now documents that it may return `null` for an unrecognised type.

Covered by `CellTagHandlerTest`.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/v07/handlers/CellTagHandler.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/v07/handlers/CellTagHandler.java
@@ -33,6 +33,7 @@ import org.apache.fesod.sheet.constant.ExcelXmlConstants;
 import org.apache.fesod.sheet.constant.FesodSheetConstants;
 import org.apache.fesod.sheet.context.xlsx.XlsxReadContext;
 import org.apache.fesod.sheet.enums.CellDataTypeEnum;
+import org.apache.fesod.sheet.exception.ExcelAnalysisException;
 import org.apache.fesod.sheet.metadata.GlobalConfiguration;
 import org.apache.fesod.sheet.metadata.data.ReadCellData;
 import org.apache.fesod.sheet.read.metadata.holder.xlsx.XlsxReadSheetHolder;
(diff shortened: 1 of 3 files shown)


delei (approved):
LGTM

PR #972, opened 2026-07-27
fix: copy default read converter map to isolate per-reader custom converters

### LINKED ISSUE #971: [Bug] Read-side custom converter leaks into the shared global converter map
### Search before asking

### Fesod version

current main (1f03942)

### JDK version

### Operating system

Linux (not OS-specific)

### Steps To Reproduce

A converter registered on one `ExcelReader` permanently pollutes the process-wide default converter registry, so a later, unrelated read (with no converter registered) still applies it.

```java
@Data
public class StringRow {
    @ExcelProperty("value")
    private String value;
}

// A converter that appends a marker so leakage is observable.
public class MarkerConverter implements Converter<String> {
    @Override public Class<?> supportJavaTypeKey() { return String.class; }
    @Override public CellDataTypeEnum supportExcelTypeKey() { return CellDataTypeEnum.STRING; }
    @Override
    public String convertToJavaData(ReadCellData<?> cellData, ExcelContentProperty p, GlobalConfiguration g) {
        return cellData.getStringValue() + " [MARKER]";
    }
    @Override
    public WriteCellData<?> convertToExcelData(String value, ExcelContentProperty p, GlobalConfiguration g) {
        return new WriteCellData<>(value);
    }
}

File file = ...; // an xlsx whose single data cell holds "hello"

// First read: register the marker converter.
List<StringRow> first = new ArrayList<>();
FesodSheet.read(file, StringRow.class, new PageReadListener<>(first::addAll))
        .registerConverter(new MarkerConverter())
        .sheet().doRead();

// Second read: fresh reader, NO converter registered.
List<StringRow> second = new ArrayList<>();
FesodSheet.read(file, StringRow.class, new PageReadListener<>(second::addAll))
        .sheet().doRead();
```

### Current Behavior

`first` correctly contains `hello [MARKER]`, but `second` also contains `hello [MARKER]` — the converter registered only on the first read is still applied.

### Expected Behavior

`second` should contain `hello` (no marker). A converter registered on one reader must not affect any other reader.

### Anything else?

Root cause: the workbook-level read holder aliases the shared static default converter map instead of copying it:

https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/holder/AbstractReadHolder.java#L122-L126

```java
if (parentAbstractReadHolder == null) {
    setConverterMap(DefaultConverterLoader.loadDefaultReadConverter()); // shared static map, no copy
} else {
    setConverterMap(new HashMap<>(parentAbstractReadHolder.getConverterMap()));
}
```

`DefaultConverterLoader.loadDefaultReadConverter()` returns the static `allConverter` map by reference. The block immediately below then does `getConverterMap().put(...)` for each registered custom converter, mutating that shared static map, so the registration survives the reader and leaks into every later read on the same JVM.

The write side already handles this correctly by copying (`AbstractWriteHolder.java:271`):

```java
setConverterMap(new HashMap<>(DefaultConverterLoader.loadDefaultWriteConverter()));
```

Doing the same on the read side fixes it:

```java
setConverterMap(new HashMap<>(DefaultConverterLoader.loadDefaultReadConverter()));
```

Aias00:
## Summary

Fixes #971.

The workbook-level read holder aliased `DefaultConverterLoader`'s shared static `allConverter` map instead of copying it, so custom converters registered via `registerConverter(...)` were `put()` into the global map and leaked into every later, unrelated read on the same JVM. This mirrors the write side (`AbstractWriteHolder:271`), which already copies.

## Root cause

`AbstractReadHolder`:

```java
if (parentAbstractReadHolder == null) {
    setConverterMap(DefaultConverterLoader.loadDefaultReadConverter()); // shared static, no copy
} else {
    setConverterMap(new HashMap<>(parentAbstractReadHolder.getConverterMap()));
}
```

`loadDefaultReadConverter()` returns the static `allConverter` map by reference, so the `getConverterMap().put(...)` loop right below mutates the shared map and the registration survives the reader.

## Fix

```java
setConverterMap(new HashMap<>(DefaultConverterLoader.loadDefaultReadConverter()));
```

## Verification

Added `ReadConverterIsolationTest` which:

1. reads a file with a registered marker converter → values carry the marker (expected);
2. reads the same file with a **fresh reader and no converter** → asserts values carry **no** marker.

Before the fix, step 2 failed (`hello [MARKER]` instead of `hello`). After the fix it passes. Full `fesod-sheet` suite: `Tests run: 668, Failures: 0, Errors: 0`.

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/DefaultConverterLoader.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/DefaultConverterLoader.java
@@ -25,6 +25,8 @@
 
 package org.apache.fesod.sheet.converters;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import org.apache.fesod.common.util.MapUtils;
 import org.apache.fesod.sheet.converters.ConverterKeyBuild.ConverterKey;
@@ -139,6 +141,7 @@ public class DefaultConverterLoader {
         putAllConverter(new StringNumberConverter());
         putAllConverter(new StringStringConverter());
         putAllConverter(new StringErrorConverter());
+        allConverter = Collections.unmodifiableMap(allConverter);
     }
 
     private static void initDefaultWriteConverter() {
@@ -176,6 +179,7 @@ public class DefaultConverterLoader {
         putWriteStringConverter(new LongStringConverter());
         putWriteStringConverter(new ShortStringConverter());
         putWriteStringConverter(new StringStringConverter());
+        defaultWriteConverter = Collections.unmodifiableMap(defaultWriteConverter);
     }
 
     /**
@@ -185,6 +189,15 @@ public class DefaultConverterLoader {
      */
     public static Map<ConverterKey, Converter<?>> loadDefaultWriteConverter() {
         return defaultWriteConverter;
+    }
+
+    /**
+     * Copy default write converter
+     *
+     * @return
+     */
+    public static Map<ConverterKey, Converter<?>> copyDefaultWriteConverter() {
+        return new HashMap<>(loadDefaultWriteConverter());
     }
 
     private static void putWriteConverter(Converter<?> converter) {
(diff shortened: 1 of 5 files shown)


bengbengbalabalabeng:
Is it possible to have the `load*` methods in `DefaultConverterLoader` return an immutable `Map`, and additionally provide `copy*` methods for creating new mutable copies? This would reduce the risk of misuse during subsequent maintenance.

psxjoy (approved):

bengbengbalabalabeng (approved):
LGTM

PR #977, opened 2026-07-27
fix: remove indexFieldMap entry by field key, not running counter

### LINKED ISSUE #976: [Bug] ClassUtils removes the wrong indexFieldMap entry when excluding a column
### Search before asking

### Fesod version

current main (50a6b5b)

### JDK version

### Operating system

Linux (not OS-specific)

### Steps To Reproduce

When a `WriteHolder` excludes a column and the bean has fields with explicit `@ExcelProperty(index = N)`, the wrong entry is removed from the internal `indexFieldMap`, dropping an unrelated explicit-index field.

```java
private static class ComplexEntity {
    @ExcelProperty(index = 0)
    private String id;        // explicit index 0 -> indexFieldMap[0]

    @ExcelProperty(index = 2)
    private String name;      // explicit index 2 -> indexFieldMap[2]

    @ExcelProperty(order = 10)
    private String email;     // no explicit index -> NOT in indexFieldMap

    private String noAnnotationField;
}

// exclude "email" (a field WITHOUT an explicit index)
WriteHolder writeHolder = ...; // excludeColumnFieldNames = {"email"}

FieldCache fieldCache = ClassUtils.declaredFields(ComplexEntity.class, writeHolder);
Map<Integer, FieldWrapper> indexFieldMap = fieldCache.getIndexFieldMap();

// expected: id and name remain (email was never in the map)
// actual: id (index 0) is gone; indexFieldMap == {2: name}
```

### Current Behavior

After excluding `email`, `indexFieldMap` is `{2: name}` — the `id` entry (explicit index 0) has been wrongly removed, even though `id` was not excluded.

### Expected Behavior

`indexFieldMap` should be `{0: id, 2: name}`. Excluding a field that has no explicit index must not perturb the explicit-index map.

### Anything else?

Root cause: in the exclude/include loop, the ignored branch removes from `indexFieldMap` using the running counter `index` instead of the field's own key:

https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ClassUtils.java#L364-L368

```java
if (writeHolder.ignore(field.getFieldName(), entry.getKey())) {
    ignoreSet.add(field.getFieldName());
    indexFieldMap.remove(index);   // BUG: 'index' is the running counter, not the field's key
} else {
    ...
}
```

`indexFieldMap` is keyed by the field's explicit `@ExcelProperty(index = N)` value (see `declaredOneField`), and for explicit-index fields the `sortedFieldMap` position `entry.getKey()` equals that explicit index (see `buildSortedAllFieldMap`). So the ignored field's own entry is keyed by `entry.getKey()`, not by the running counter. Removing by the counter drops whichever explicit-index entry happens to match the counter and leaves the ignored field's entry in place when their indices differ.

The fix is to remove by the field's own key:

```java
indexFieldMap.remove(key);
```

For a field with an explicit index, `key == entry.getKey()` is its `indexFieldMap` key, so its own entry is removed; for a field without an explicit index, `key` is not in `indexFieldMap`, so the call is a no-op (correct — it was never there).

The corruption is observable downstream: `ExcelHeadProperty.initColumnProperties` passes `indexFieldMap.containsKey(entry.getKey())` as `forceIndex` into each `Head`, so an excluded unrelated field flips a sibling column's `forceIndex` from `true` to `false`, which `DefaultAnalysisEventProcessor` reads to drive head-to-column matching.

Aias00:
## Summary

Fixes #976.

When a `WriteHolder` excludes a column, the ignored branch removed from `indexFieldMap` using the running loop counter `index` instead of the field's own `sortedFieldMap` key. For explicit-index fields, that key (`entry.getKey()`) equals the field's `@ExcelProperty(index)` — its `indexFieldMap` key — so removing by the counter dropped an unrelated explicit-index entry and left the ignored field's entry in place.

## Root cause

`ClassUtils.doDeclaredFields`:

```java
if (writeHolder.ignore(field.getFieldName(), entry.getKey())) {
    ignoreSet.add(field.getFieldName());
    indexFieldMap.remove(index);   // BUG: 'index' is the running counter, not the field's key
} else {
    ...
}
```

`indexFieldMap` is keyed by the field's explicit `@ExcelProperty(index = N)` (see `declaredOneField`), and for explicit-index fields the `sortedFieldMap` position `entry.getKey()` equals that index (see `buildSortedAllFieldMap`). So the ignored field's own entry is keyed by `entry.getKey()`, not the counter.

## Fix

```java
indexFieldMap.remove(key);
```

For a field with an explicit index, `key == entry.getKey()` is its `indexFieldMap` key, so its own entry is removed; for a field without an explicit index, `key` is not in `indexFieldMap`, so the call is a no-op (correct — it was never there).

The corruption is observable downstream: `ExcelHeadProperty.initColumnProperties` passes `indexFieldMap.containsKey(entry.getKey())` as `forceIndex` into each `Head`, which `DefaultAnalysisEventProcessor` reads to drive head-to-column matching.

## Verification

Added `ClassUtilsTest.test_declaredFields_WriteHolder_exclude_preservesUnrelatedExplicitIndex`: with `ComplexEntity` (`id` index 0, `name` index 2, `email` order 10), excluding `email` must keep `indexFieldMap` as `{0: id, 2: name}`. Before the fix `id` (index 0) was wrongly removed; after the fix it stays. Full `fesod-sheet` suite: `Tests run: 668, Failures: 0`.

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ClassUtils.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ClassUtils.java
@@ -364,7 +364,11 @@ public class ClassUtils {
             // The current field needs to be ignored
             if (writeHolder.ignore(field.getFieldName(), entry.getKey())) {
                 ignoreSet.add(field.getFieldName());
-                indexFieldMap.remove(index);
+                // indexFieldMap is keyed by the field's explicit @ExcelProperty(index), which for
+                // explicit-index fields equals the sortedFieldMap position (entry.getKey()); remove
+                // by that key, not the running counter, otherwise an unrelated explicit-index entry
+                // is dropped and the ignored field's entry may survive.
+                indexFieldMap.remove(key);
             } else {
                 // Mandatory sorted fields
                 if (indexFieldMap.containsKey(key)) {
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/util/ClassUtilsTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/util/ClassUtilsTest.java
@@ -203,6 +203,29 @@ class ClassUtilsTest {
     }
 
     @Test
+    void test_declaredFields_WriteHolder_exclude_preservesUnrelatedExplicitIndex() {
+        // Excluding a field that has NO explicit @ExcelProperty(index) (here "email",
+        // which uses order=10) must not perturb indexFieldMap, which only holds fields
+        // that DO carry an explicit index (ComplexEntity: id->0, name->2).
+        Mockito.when(globalConfiguration.getFiledCacheLocation()).thenReturn(CacheLocationEnum.NONE);
+
+        Mockito.when(writeHolder.excludeColumnFieldNames()).thenReturn(Collections.singleton("email"));
+        Mockito.when(writeHolder.ignore(Mockito.anyString(), Mockito.anyInt())).thenReturn(false);
+        Mockito.when(writeHolder.ignore(Mockito.eq("email"), Mockito.anyInt())).thenReturn(true);
+
+        FieldCache fieldCache = ClassUtils.declaredFields(ComplexEntity.class, writeHolder);
+
+        Map<Integer, FieldWrapper> indexFieldMap = fieldCache.getIndexFieldMap();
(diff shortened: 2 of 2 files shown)


delei (approved):
LGTM

Issue #696, opened 2025-11-19
[Bug] 增对特殊字符可以进行处理吗


xfn-jjw:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar.

### Fesod version

1.0.0

### JDK version

1.8

### Steps To Reproduce

<img width="893" height="171" alt="Image" src="https://github.com/user-attachments/assets/09f47fdf-3710-4d52-9cec-e63b624619d1" />数据是 一个不可见字符 \u0002 导出到excel乱码为 ？ 号

### Current Behavior

<img width="197" height="51" alt="Image" src="https://github.com/user-attachments/assets/e5ab82f7-650d-4a18-bf59-950ba3a0f5a3" /> 实际的数据是 FZ85962-Black-XLMQR

### Expected Behavior

预期导出导入都与原数据一致

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xfn-jjw:
<img width="309" height="81" alt="Image" src="https://github.com/user-attachments/assets/7393dcc0-db07-414c-b2eb-0c1eb0ee1541" />实际的数据

psxjoy:
A simple demo is a better choice to make someone helped.

ongdisheng:
## Reproduction Steps

I created a test case to reproduce this issue:
```java
public class ControlCharacterReproduceTest {

    @Test
    public void reproduceIssue696() {
        // Write data with control character \u0002
        List<DemoData> writeData = new ArrayList<>();
        DemoData data = new DemoData();
        data.setString("Product\u0002Code");
        data.setDate(new Date());
        data.setDoubleData(123.45);
        writeData.add(data);

        File file = new File("issue696_reproduce.xlsx");
        FesodSheet.write(file, DemoData.class).sheet("Sheet1").doWrite(writeData);

        // Read back
        List<DemoData> readData = new ArrayList<>();
        FesodSheet.read(file, DemoData.class, new AnalysisEventListener<DemoData>() {
            @Override
            public void invoke(DemoData data, AnalysisContext context) {
                readData.add(data);
            }

            @Override
            public void doAfterAllAnalysed(AnalysisContext context) {
            }
        }).sheet().doRead();

        String original = writeData.get(0).getString();
        String result = readData.get(0).getString();

        System.out.println("Original: " + original);
        System.out.println("Result:   " + result);
    }
}

```

## Result

[...]

PR #991, opened 2026-08-06
fix: decode _xHHHH_ escapes when reading inline string cells

nkuprins:
Related: #696 

Characters that XML 1.0 forbids are stored in a cell as `_xHHHH_` escapes. Fesod undid them only for cells backed by `sharedStrings.xml`, so `t="inlineStr"` and `t="str"` reached the caller with the raw escape. A value written by Fesod did not survive being read back by Fesod. The clearest case is a value escaped with `EscapeHexCellWriteHandler`, which stores the literal `_xB9f0_` as `_x005F_xB9f0_` precisely so a decoding reader restores it:

| reading a cell written as `Product_x005F_xB9f0_Code` | result |
| --- | --- |
| POI | `Product_xB9f0_Code` |
| Fesod, before / after | `Product_x005F_xB9f0_Code` / `Product_xB9f0_Code` |

Why this is a defect and not intended behaviour:
- Decoding is already an asserted contract for the other storage form - `CompatibilityTest#readXlsxWithEscapeSequence` pins it for `sharedStrings.xml`.
- POI decodes inline strings on both of its read paths - the DOM `XSSFCell` and the streaming `XSSFSheetXMLHandler`, each via `XSSFRichTextString#getString()`.
- No test asserts that a raw escape survives a read.

## What's changed?

- **`XlsxEscapeUtils` (new)** - `utfDecode` moved here from `SharedStringsTableHandler` unchanged, and both read paths now call it. The smaller change would have been to make the existing method `public` and call it from `CellTagHandler`, but that leaves a cell parser depending on the `sharedStrings.xml` parser for decoding that has nothing to do with shared strings, and the next caller inherits the same detour. The escape convention belongs to neither handler, so it moved to `util/`. Happy to switch to the two-line version if you prefer the smaller diff.
- **`CellTagHandler`** - the `DIRECT_STRING` branch now decodes instead of falling through to the `ERROR` branch; it sets the same `STRING` type as before.
- **`HexEscapeRoundTripTest` (new)** - one round trip through the handler: `Product_x0002_Code` written with `EscapeHexCellWriteHandler` is stored as `Product_x005F_x0002_Code` and must read back as the literal `Product_x0002_Code`.

## Checklist


fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/v07/handlers/CellTagHandler.java
```diff
@@ -116,6 +117,10 @@ public void endElement(XlsxReadContext xlsxReadContext, String name) {
                 tempCellData.setStringValue(stringValue);
                 break;
             case DIRECT_STRING:
+                // Undo the '_xHHHH_' escapes of characters XML forbids
```
alaahong, line 110: The reader now decodes `_xHHHH_`, but the default writer (SXSSF inline strings) never escapes them. Verified locally: writing `Product_x0002_Code` with the default writer and reading it back now returns `Product\u0002Code` (STX control char) instead of the literal — before this change it returned the literal. For third-party files this is the correct fix for #696, but for fesod→fesod round trips it's a silent behaviour change. Suggest aligning the writer (escape by default, like `EscapeHexCellWriteHandler`) or documenting that literal `_xHHHH_` text requires registering that handler.

fesod-sheet/src/test/java/org/apache/fesod/sheet/readwrite/HexEscapeRoundTripTest.java
```diff
@@ -0,0 +1,60 @@
...
+        data.setName("Product_x0002_Code");
+        File file = createTempFile("hex-escape", ExcelFormat.XLSX);
+        FesodSheet.write(file, SimpleData.class).sheet().doWrite(Collections.singletonList(data));
+
+        CollectingReadListener<SimpleData> listener = new CollectingReadListener<>();
+        FesodSheet.read(file, SimpleData.class, listener).sheet().doRead();
+        List<SimpleData> rows = listener.getRows();
+
+        Assertions.assertEquals(1, rows.size());
+        Assertions.assertEquals("Product" + STX + "Code", rows.get(0).getName());
```
alaahong: This assertion only holds because the default writer does not escape `_xHHHH_`. If the writer is fixed to escape (spec-correct), the same input reads back as the literal `Product_x0002_Code` and this test would fail for the wrong reason. Suggest writing via `EscapeHexCellWriteHandler` (as in #1001's round-trip test) or using a POI/Excel-generated fixture containing real escapes.

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/XlsxEscapeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+package org.apache.fesod.sheet.util;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Escapes defined by the Office Open XML spec, section 3.18.9.
+ *
+ *
+ */
```
alaahong, line 45: Since `utfDecode` is now a shared public API, consider adding a parameterized unit test (edge cases: adjacent escapes, `_x000D_`, uppercase `_X`, truncated sequences, non-hex digits, pattern at end of string) and expanding the class-level javadoc (purpose, the two call sites, OOXML §3.18.9).

alaahong (commented):
Nice cleanup — moving `utfDecode` to `util/` so `CellTagHandler` doesn't depend on the shared-strings parser is the right call, and `t="str"` is covered since it maps to `DIRECT_STRING`. Comments inline. One more note: the escape convention is now implemented twice (`XlsxEscapeUtils` for reading, `EscapeHexCellWriteHandler` for writing); consider cross-referencing them or unifying in `XlsxEscapeUtils` to prevent silent drift.

fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/v07/handlers/CellTagHandler.java
nkuprins: Hi @alaahong,

I see, but if I take the first option (escape by default), then do you agree that we need an opt-out - `escapeHexText(false)` on the write? Without it, escapes that arrived from another producer get escaped again on the way in, with no way to write them through unchanged. 

Also, probably escape by default isn't a small addition to this PR, so not sure if I should do it here.

Thank you for the review!

nkuprins: I chose to add javadoc note instead, as suggested in the second option. This probably belongs on the website too, but I'm not sure which section - FAQ?

alaahong (approved):
LGTM

fesod-sheet/src/test/java/org/apache/fesod/sheet/util/XlsxEscapeUtilsTest.java
```diff
@@ -0,0 +1,106 @@
...
+     * POI is the reference for this decoding, so it is also the oracle: any input where the two disagree is a defect
+     * here, whatever the table above says.
+     */
+    @ParameterizedTest(name = "[{index}] {0} decodes as POI does")
+    @MethodSource
+    void matchesPoiForEveryInput(String input) {
+        Assertions.assertEquals(new XSSFRichTextString(input).getString(), XlsxEscapeUtils.utfDecode(input));
+    }
+
+    static Stream<String> matchesPoiForEveryInput() {
```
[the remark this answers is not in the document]
nkuprins: The source is the union of `decodesEscapes()` and `leavesTextWithoutAnEscapeAlone()`, so "every
input" means every input in this class

fesod-sheet/src/test/java/org/apache/fesod/sheet/readwrite/HexEscapeRoundTripTest.java
```diff
@@ -0,0 +1,67 @@
...
+                .registerWriteHandler(new EscapeHexCellWriteHandler())
+                .sheet()
+                .doWrite(Collections.singletonList(data));
+
+        CollectingReadListener<SimpleData> listener = new CollectingReadListener<>();
+        FesodSheet.read(file, SimpleData.class, listener).sheet().doRead();
+        List<SimpleData> rows = listener.getRows();
+
+        Assertions.assertEquals(1, rows.size());
+        Assertions.assertEquals("Product_x0002_Code", rows.get(0).getName());
```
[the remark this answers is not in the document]
nkuprins: PR description changed

fesod-sheet/src/test/java/org/apache/fesod/sheet/util/XlsxEscapeUtilsTest.java
```diff
@@ -0,0 +1,106 @@
...
+                Arguments.of("_x0041_", "A"),
+                Arguments.of("_x0041_tail", "Atail"),
+                Arguments.of("head_x0041_", "headA"),
+                Arguments.of("head_x0041_tail", "headAtail"),
+                Arguments.of("_x0041__x0042_", "AB"),
+                Arguments.of("__x0041_", "_A"),
+                Arguments.of("_x000D_", "\r"),
+                // the hex digits are case insensitive, the leading x is not - see leavesTextWithoutAnEscapeAlone
+                Arguments.of("_x00e9_", "é"),
+                Arguments.of("_x00E9_", "é"),
```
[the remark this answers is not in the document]
nkuprins: Doesn't seem to be a problem - non-ASCII literals are already used in `CompatibilityTest`, etc.

Issue #1000, opened 2026-08-09
[Enhancement] Improve test coverage and cleanup for EscapeHexCellWriteHandler


nkuprins:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar.

### Motivation

While working on #991, I noticed that `EscapeHexCellWriteHandler` has no dedicated tests, and the pattern-scanning code has some edges worth tidying.

### Solution

I will do it in 2 PR.

### 1. Testing
- [x] Cover the escaping logic itself - valid patterns get rewritten, invalid/edge-case ones don't,
  and non-applicable inputs (wrong cell data type, null values) are safely ignored.
- [x] Cover it end-to-end - write a file through Fesod with the handler registered and confirm the
  escaped value survives being read back.

### 2. Refactoring/cleanup
> Hold off reviewing this section until 1. is merged - the tests need to pin current behaviour before
> anything underneath them moves.

- [x] Reduce duplication in the pattern-scanning code - the match-detection logic is currently
  written out twice across two methods.
- [x] Improve readability of the low-level string offsets (magic numbers, member ordering), so the
  logic is easier to verify by inspection.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

delei:
The unit test class might have been omitted when renaming the package. You can refer to the PR #470 and issue #364.

nkuprins:
> The unit test class might have been omitted when renaming the package. You can refer to the PR [#470](https://github.com/apache/fesod/pull/470) and issue [#364](https://github.com/apache/fesod/issues/364).

Confirmed! The accidental deletion happened in #770. Then, I guess I will restore it and see if something can be improved there.

PR #1001, opened 2026-08-09
test: add tests for EscapeHexCellWriteHandler

nkuprins:
Related: #1000

`EscapeHexCellWriteHandler` has no tests. It used to have them, but they were deleted by accident (read #1000).

 _This PR is not a blind restore of those lost tests! I took a different approach; for example,  `@ParameterizedTest` were used for better scalability and readability, etc._

## What's changed?
 
 - **Two parameterised tables** covering the escaping rule itself. Seven inputs that must be rewritten (`_xHHHH_` to `_x005F_xHHHH_`) and eight that must be left exactly as typed.
- **Non-idempotency**: an already-escaped literal is escaped *again*.
- **Guard tests** for the inputs the handler must not act on: non-`STRING` cell data, a null `WriteCellData`, and a `STRING` holding a null value.
- **One end-to-end case**, parameterised over XLSX/XLS/CSV with `@ExcelFormatSource`: write a file through Fesod with the handler registered, read it back, and require the caller's literal. It sits in its own class tagged `ROUND_TRIP`, which keeps the rest tagged `UNIT`.

## Checklist

### DIFF
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/write/handler/EscapeHexCellWriteHandlerRoundTripTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/write/handler/EscapeHexCellWriteHandlerRoundTripTest.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fesod.sheet.write.handler;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.apache.fesod.sheet.FesodSheet;
+import org.apache.fesod.sheet.testkit.Tags;
+import org.apache.fesod.sheet.testkit.base.AbstractExcelTest;
+import org.apache.fesod.sheet.testkit.enums.ExcelFormat;
+import org.apache.fesod.sheet.testkit.params.ExcelFormatSource;
+import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.ss.usermodel.WorkbookFactory;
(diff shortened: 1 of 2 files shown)


delei:
LGTM.
@alaahong PTAL when it is convenient for you.

fesod-sheet/src/test/java/org/apache/fesod/sheet/write/handler/EscapeHexCellWriteHandlerRoundTripTest.java
```diff
@@ -0,0 +1,83 @@
...
+                .head(Collections.singletonList(Collections.singletonList("value")))
+                .registerWriteHandler(new EscapeHexCellWriteHandler())
+                .sheet("escape")
+                .doWrite(rows);
+        return file;
+    }
+
+    private String readBackFirstDataValue(File file, ExcelFormat format) throws IOException {
+        if (format == ExcelFormat.CSV) {
+            try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
```
alaahong: why not Files.newBufferedReader(file.toPath(), StandardCharsets.UTF_8) ? Avoid to depend system encoding.

fesod-sheet/src/test/java/org/apache/fesod/sheet/write/handler/EscapeHexCellWriteHandlerTest.java
```diff
@@ -0,0 +1,118 @@
...
+            SXSSFCell cell = workbook.createSheet().createRow(0).createCell(0);
+            WriteCellData<?> emptyStringData = new WriteCellData<>(CellDataTypeEnum.STRING);
+
+            Assertions.assertDoesNotThrow(
+                    () -> handler.afterCellDataConverted(null, null, null, cell, null, 0, Boolean.FALSE));
+            Assertions.assertDoesNotThrow(
+                    () -> handler.afterCellDataConverted(null, null, emptyStringData, cell, null, 0, Boolean.FALSE));
+            Assertions.assertNull(emptyStringData.getStringValue());
+        }
+    }
```
alaahong, line 123: It should be fine, a little suggestion, seems it's bringing some new additional exceptions by testing code? 
https://github.com/apache/fesod/actions/runs/31603043137/job/94135710567

Image

alaahong (changes requested):
Generally fine, can you revise the code as common practice?

fesod-sheet/src/test/java/org/apache/fesod/sheet/write/handler/EscapeHexCellWriteHandlerTest.java
nkuprins: Hi @alaahong! Thank you for the review! 

I partly based my setup on `PoiUtilsTest:test_customHeight_unsupportedType_false`, which has exactly the same `WARN`. This is why I overlooked it. 

One fix could be `sheet.flushRows();`. However, in my case, the handler does nothing with the cell beyond `cell instanceof SXSSFCell`, so I decided to fix it with `Mockito.mock(SXSSFCell.class)`, and it also makes the code easier to read. 

The one thing the mock gave up was coverage of the `instanceof` check itself, so I've pushed a follow-up test, `afterCellDataConverted_ignoresNonSxssfCells`, which runs the handler over a mocked `HSSFCell` and asserts the value is left untouched.

Please resolve the conversation if you accept this fix :)

alaahong (approved):

PR #1002, opened 2026-08-10
refactor: make objects immutable and remove redundant methods

bengbengbalabalabeng:

As title. 

## What's changed?

- Remove some unused Lombok annotations.
- Refactor these classes with only read-only properties into immutable objects.

## Checklist

- [ ] I have written the necessary doc or comment.


alaahong:
generally good, do we have the possible to mark as @Deprecated for setter and propose to remove since a future version to avoid breaking change on existing usage?
And meanwhile ,for the unit test, can we add more dedicated case on the immutable?

PR #1003, opened 2026-08-10
refactor: simplify hex escape pattern scanning

nkuprins:
Related: #1000 

The second of the two PRs. This one does the refactor/cleanup. No behaviour change.

## What's changed?

- **The match loop is written once instead of twice.** The scan used to live in two methods: `escapeHex` located the
  first valid pattern, then `processWithPatterns` located every later one. Both carried the same match-handling code - bounds check, closing-`_` test, hex validation, append, so any change to the rule had to be made in two places. It is now a single loop.
- **The no-match path allocates nothing.** The `StringBuilder` is created on the first match rather than up front, so a cell containing no `_xHHHH_`  - costs one `indexOf` scan and no allocation.
- **The offsets have names.** Each iteration derives `hexStart`, `suffixIndex` and `patternEnd` from `PREFIX_LENGTH` and `HEX_DIGIT_COUNT`.
- **`isValidHexFast` is now `isHexDigits`**, and the lookup table has a note on why it is not `Character.digit(c, 16)` - that method also accepts non-ASCII digits such as `U+0663`, which OOXML never encodes.

## Checklist


fesod-sheet/src/main/java/org/apache/fesod/sheet/write/handler/EscapeHexCellWriteHandler.java
```diff
@@ -57,101 +71,56 @@ public void afterCellDataConverted(
...
-
-                // Append content between patterns
+            if (originalString.charAt(suffixIndex) == '_' && isHexDigits(originalString, hexStart)) {
+                if (result == null) {
+                    result = new StringBuilder(length + 64);
+                }
+                // Append content since the previous match, then the escaped pattern
                 result.append(originalString, lastEnd, patternIndex);
-                // Append escaped pattern
                 result.append("_x005F_x");
```
delei: How about changing the magic string `_x` or `_x005F_x` to a character constant, to eliminate the magic string?

nkuprins: I was actually thinking about this but was not sure. Now will do.
\
In that case, what about `SUFFIX = '_';` for `result.append('_');`?

delei: IMO, no need to handle `_` characters

delei (approved):
LGTM

PR #1007, opened 2026-08-11
fix: prioritize subclass fields and ignore superclass annotations on field shadowing

bengbengbalabalabeng:

Similar to #427

Fix the issue where a subclass attribute with the same name (field shadowing) in a class inheritance chain mistakenly inherits the parent class annotation (such as `@DateTimeFormat`), causing abnormal sheet export formatting.

*Special thanks to @nkuprins for identifying this issue in Discussion #919 !*

## What's changed?

Refactored the field collection logic in `ClassUtils#doDeclaredFieldContentMap` and extracted it into a reusable `FieldUtils#resolveFields` utility method.

Scope of affected annotations:
- `@ContentStyle`
- `@ContentFontStyle`
- `@ExcelProperty` (converter attribute)
- `@DateTimeFormat`
- `@NumberFormat`

## Checklist


fesod-sheet/src/main/java/org/apache/fesod/sheet/util/FieldUtils.java
```diff
@@ -178,4 +186,33 @@ public static Field getField(final Class<?> cls, final String fieldName, final b
...
+
+    /**
+     * Resolves and retrieves all declared fields from the specified class and its inheritance hierarchy.
+     * If a field with the same name (resolved by CGLIB) is declared in both a subclass and a superclass,
+     * the subclass field definition takes precedence.
+     *
+     * @param cls the target {@link Class}, must not be {@code null}
+     * @return a {@link List} containing all resolved fields, or an empty list if no fields are found
+     */
+    public static List<Field> resolveFields(Class<?> cls) {
```
delei: How about this method name?

```suggestion
    public static List<Field> collectResolvedFields(Class<?> cls) {
```

<details>

<summary>AI Assistant</summary>

I think `resolveFields` is a bit too generic for what this method actually does.

The implementation is not only resolving fields, but also walking the inheritance hierarchy, filtering out static/synthetic fields, and applying shadowing precedence so that subclass fields win over superclass fields.

To make that behavior clearer, I’d recommend renaming it to `collectResolvedFields`. This name better communicates that the method is collecting the final resolved field set after inheritance and shadowing are taken into account.

</details>

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/FieldUtils.java
```diff
@@ -178,4 +186,33 @@ public static Field getField(final Class<?> cls, final String fieldName, final b
...
+                        || field.isSynthetic()
+                        || !fieldNames.add(resolveCglibFieldName(field));
+                if (shouldSkip) {
+                    continue;
+                }
+                result.add(field);
+            }
+            tempClass = tempClass.getSuperclass();
+        }
+        return Collections.unmodifiableList(result);
```
delei: As a public utility method, I suggest returning result directly instead of Collections.unmodifiableList(result).
Otherwise, callers may hit UnsupportedOperationException during common follow-up operations (e.g., add / remove / sort), which can be inconsistent with typical utility-method expectations.

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/FieldUtils.java
bengbengbalabalabeng: I think having `collect` and `resolved` together is a bit redundant.

How about `resolveAllFields`? (Reference: `FieldUtils#getAllFieldsList` in Apache Commons Lang).

delei: `resolveAllFields` is good.

delei (approved):
LGTM

PR #1012, opened 2026-08-13
docs(website): unify example code blocks to JEP-512 style

bengbengbalabalabeng:

Unify the code style of all example code blocks in the documentation to follow a consistent `JEP-512` (Compact Source Files and Instance Main Methods) pattern, replacing the previously varied `@Test` method styles.

## What's changed?

- Before:

```java
// Style 1
public static void main(String[] args) {
    // .....
}

// Style 2
@Test
void readAndWrite() {
    // ......
}
```

- After:

```java
void main() {
    // ......
}
```

## Checklist


website/docs/sheet/write/pojo.md
```diff
@@ -204,12 +201,11 @@ public void includeColumnOrderWrite() {
 Add `.orderByIncludeColumn(true)` to follow the collection's order instead:
 
 ```java
-@Test
-public void orderByIncludeColumnWrite() {
-    String fileName = "includeColumnFieldWrite" + System.currentTimeMillis() + ".xlsx";
+void main() {
```
alaahong, line 204: will the replacement making confuse with psvm ?

bengbengbalabalabeng: Thank you for the reminder! The purpose of switching to the `JEP‑512` style this time is mainly to keep the example code in the documentation structurally consistent, and it is not intended to imply any connection between these examples and `PSVM`.

In the updated example, `void main()` is used only to improve readability; it does not represent the actual entry-point method, nor does it alter the semantics of the example. Therefore, it will not be confused with `PSVM`.

alaahong:
Generally good, only on concern is the sample only support on JDK 25+, but current major user are from java 8 - 21...any confuse while following guide to code?

delei:
> Generally good, only on concern is the sample only support on JDK 25+, but current major user are from java 8 - 21...any confuse while following guide to code?

Let’s keep up with the times. LOL.

bengbengbalabalabeng:
Or provide a supplementary note: indicate that the `void main()` pattern is used only to simplify and unify the example code style, and does not mean that this feature supports only Java 25+ versions.

Issue #950, opened 2026-07-15
[Task] add support for column index limit in ReadSheet


delei:
### Motivation

add support for reading a sheet based on rows and columns index.

### How to complete

#942 

### Task List

- [x] Introducing a new parameter when reading the file
- [x] Supports the xlsx file format
- [x] Supports the csv file format
- [x] Supports the xls file format
- [x] Complete the documentation for the new parameters, including both Chinese and English versions

delei:
Hi, @sapienza88 

Thank you for your contribution. You can continue to submit a new PR to link to this issue.

bengbengbalabalabeng:
Hi, @sapienza88 

Thanks again for your contributions to this issue.
We’re now down to the final task related to **"completing the documentation for the new parameters (both English and Chinese versions)"**.
Before proceeding, I’d like to check whether you’re interested in continuing with this last part.
If you prefer not to take it, no problem, we can help move it forward.
Just let me know your preference.

sapienza88:
@bengbengbalabalabeng yes I can do that

sapienza88:
@bengbengbalabalabeng @delei I am adding reference to the PR for the docs https://github.com/apache/fesod/pull/1008, thank you for reviewing it.

PR #1015, opened 2026-08-15
refactor: replace includeColumnIndexes list lookup with ColumnIndexResolver

bengbengbalabalabeng:

Related: #950 

Replace duplicate `List<Integer> includeColumnIndexes` logic across read handlers with a strategy-based `ColumnIndexResolver`.

## What's changed?

- Removed `ReadSheet#columnIndexes` and added `ReadSheet#columnIndexResolver`.
- Added `ReadSheetHolder#determineTargetColumnIndex(columnIndex)` method.
- Added `ColumnIndexResolver.PASS_THROUGH`: Keep original column indices.
- Added `DefaultIncludedColumnIndexResolver`: Rewrite target columns’ internal indexes to consecutive (0-based).
  - Replace the original `List.indexOf()` lookup with `Map.get()` to improve lookup efficiency and semantic clarity.

## Checklist


sapienza88:
The purpose says what was done here but does not argue the benefits of the refactoring..

fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/ColumnIndexResolver.java
```diff
@@ -0,0 +1,71 @@
...
+    static ColumnIndexResolver fromInclude(List<Integer> columnIndexes) {
+        return new DefaultIncludedColumnIndexResolver(columnIndexes);
+    }
+
+    class DefaultIncludedColumnIndexResolver implements ColumnIndexResolver {
+
+        private final Map<Integer, Integer> indexMap;
+
+        DefaultIncludedColumnIndexResolver(List<Integer> columnIndexes) {
+            Validate.notEmpty(columnIndexes, "The includeColumnIndexes must not be empty");
```
sapienza88, line 56: `Validate.notEmpty(columnIndexes)` ensures the list is not null or empty, but columnIndexes could still contain a null element

fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/ColumnIndexResolver.java
```diff
@@ -0,0 +1,71 @@
...
+            List<Integer> tmpColumnIndexes = new ArrayList<>(columnIndexes);
+            this.indexMap = MapUtils.newHashMapWithExpectedSize(tmpColumnIndexes.size());
+
+            for (int targetIndex = 0; targetIndex < tmpColumnIndexes.size(); targetIndex++) {
+                this.indexMap.put(tmpColumnIndexes.get(targetIndex), targetIndex);
+            }
+        }
+
+        @Override
+        public Integer resolve(int columnIndex) {
```
sapienza88, line 68: Returning Integer (wrapper) forces autoboxing on every cell read and requires a null check, try returning primitive `int `instead.

fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/ColumnIndexResolver.java
```diff
@@ -0,0 +1,71 @@
...
+        private final Map<Integer, Integer> indexMap;
+
+        DefaultIncludedColumnIndexResolver(List<Integer> columnIndexes) {
+            Validate.notEmpty(columnIndexes, "The includeColumnIndexes must not be empty");
+
+            List<Integer> tmpColumnIndexes = new ArrayList<>(columnIndexes);
+            this.indexMap = MapUtils.newHashMapWithExpectedSize(tmpColumnIndexes.size());
+
+            for (int targetIndex = 0; targetIndex < tmpColumnIndexes.size(); targetIndex++) {
+                this.indexMap.put(tmpColumnIndexes.get(targetIndex), targetIndex);
```
sapienza88, line 63: +1 for enabling the O(1) map lookup through this.

fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/ColumnIndexResolver.java
bengbengbalabalabeng: great catch.

fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/ColumnIndexResolver.java
bengbengbalabalabeng: Thanks for the review! Regarding returning `Integer` versus primitive `int`, I suggest keeping the `Integer` return type for the following reasons:

- `DefaultIncludedColumnIndexResolver` returns existing `Integer` references directly from indexMap.
- `PASS_THROUGH` hits pre-allocated `IntegerCache` singletons for common column indexes (0–127). No runtime GC overhead.
- The downstream `ReadCellData` object itself requires an `Integer` columnIndex, so returning primitive `int` would merely defer autoboxing.
- Column filtering inherently requires a branch check to skip excluded columns. Returning `null` provides clear, type-safe semantics (`null` = filtered out) without relying on magic sentinel numbers (like -1).

delei (approved):
LGTM.

Issue #1017, opened 2026-08-16
[Task] Implement more commonly used converter classes


delei:
### Motivation

During usage, we may need to convert read or written data into specific formats. Apache Fesod (Incubating) provides a flexible converter mechanism that allows users to define custom data conversion rules to meet various business requirements. Many converter classes are already built in, but some are still missing.

### How to complete

Please link this issue in the description of the PR.

This will be a long-term task, allowing all the volunteers to divide it into multiple PRs for completion. Each PR can contain one or more converters, and each converter needs to provide the corresponding unit test code.

Only classes compatible with **JDK 8** are supported. Each converter must implement the `org.apache.fesod.sheet.converters.Converter` interface, and please leave a comment outlining your specific implementation proposal.

### Task List

The following are some conversions that are currently required
- [ ] `java.sql.Timestamp` @SyedIshmumAhnaf 
- [ ] `java.sql.Clob` @AhmedYoussef98 
- [ ] `java.time.ZonedDateTime` @alexsmolya 
- [x] `java.time.LocalTime` @leehaut 
- [ ] `java.time.OffsetDateTime` @Mikkey-f 

Please feel free to suggest any type converters you require, and please leave a comment in this issue.

**The new proposal needs to be published for at least 24 hours. If there are no objections during this period, please feel free to submit your PR at any time. These will be added to the task list and assigned to volunteers later.**

alexsmolya:
Here is an implementation proposal for `java.time.ZonedDateTime` converters, following the established `LocalDateTime` / `LocalDate` pattern:

### 1. Proposed Converter Family (`org.apache.fesod.sheet.converters.zoneddatetime`)

* **`ZonedDateTimeDateConverter`** (Write -> Excel `DATE` cell):
  * Emits `WriteCellData(value.toLocalDateTime())` with default date format (`yyyy-MM-dd HH:mm:ss`).
* **`ZonedDateTimeNumberConverter`** (Bidirectional Excel `NUMBER` serial):
  * **Read**: Parses Excel numeric serial to `LocalDateTime` (respecting `use1904windowing`), then assigns `ZoneId.systemDefault()` (`localDateTime.atZone(ZoneId.systemDefault())`).
  * **Write**: Converts `value.toLocalDateTime()` to numeric serial via `DateUtil.getExcelDate(value.toLocalDateTime(), use1904)`.
* **`ZonedDateTimeStringConverter`** (Bidirectional Excel `STRING` cell):
[...]

PR #1020, opened 2026-08-17
feat: add ZonedDateTime converters

alexsmolya:

Related: #1017

## What's changed?

Adds the approved `java.time.ZonedDateTime` converter family and registers it with the default converter loader.

- DATE and NUMBER writes use `toLocalDateTime()`, intentionally dropping zone/offset while preserving local wall-clock fields.
- Numeric reads attach `ZoneId.systemDefault()`.
- STRING conversion supports the accepted ISO and configured formatting/parsing semantics.
- Added dedicated `ZonedDateTimeConverterTest` coverage for supported directions, registration, formatting, and timezone-lossiness behavior.

## Checklist

Focused validation: 6 ZonedDateTime tests passed; Java 1.8-targeted compilation, Spotless, and `git diff --check` passed.


fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/ZonedDateTimeConverterTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.fesod.sheet.converters.zoneddatetime.ZonedDateTimeStringConverter;
+import org.apache.fesod.sheet.enums.CellDataTypeEnum;
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.ReadCellData;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.DateTimeFormatProperty;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+import org.junit.jupiter.api.Test;
+
+class ZonedDateTimeConverterTest {
```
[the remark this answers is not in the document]
alexsmolya: Tagged `ZonedDateTimeConverterTest` with `@Tag(Tags.UNIT)` following repository conventions.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeStringConverter.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    private DateTimeFormatter formatter(ExcelContentProperty contentProperty, Locale locale) {
+        if (contentProperty == null || contentProperty.getDateTimeFormatProperty() == null) {
+            return DateTimeFormatter.ISO_ZONED_DATE_TIME;
+        }
+        return DateTimeFormatter.ofPattern(
+                contentProperty.getDateTimeFormatProperty().getFormat(), locale);
+    }
+}
```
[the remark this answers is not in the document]
alexsmolya: Updated `ZonedDateTimeStringConverter` to fall back to `DateTimeFormatter.ISO_ZONED_DATE_TIME` when the format string is null or empty, and added regression test coverage in `ZonedDateTimeConverterTest`.

nkuprins (changes requested):
Consider adding a test for `use1904windowing` :)

alexsmolya:
Added focused `use1904windowing` coverage for `ZonedDateTime`, including global and field-level configuration plus write/read round-trip behavior. Targeted and related converter tests pass.

bengbengbalabalabeng:

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeDateConverter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public WriteCellData<?> convertToExcelData(
+            ZonedDateTime value, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration)
+            throws Exception {
+        LocalDateTime localDateTime = value.toLocalDateTime();
+        WriteCellData<?> cellData = new WriteCellData<>(localDateTime);
+        String format = null;
+        if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
+            format = contentProperty.getDateTimeFormatProperty().getFormat();
```
Mikkey-f, line 53: Empty `@DateTimeFormat` values are passed through to `WorkBookUtil.fillDataFormat` as-is. `fillDataFormat` only substitutes the default format when the argument is `null` — an empty string (`""`) is written into the cell style as-is, so `@DateTimeFormat(value = "")` produces a DATE cell with **no format at all** instead of falling back to `yyyy-MM-dd HH:mm:ss`.

Suggest normalizing the empty format to `null` before the call:

```java
String format = null;
if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
    format = contentProperty.getDateTimeFormatProperty().getFormat();
    if (StringUtils.isEmpty(format)) {
        format = null;
    }
}
WorkBookUtil.fillDataFormat(cellData, format, DateUtils.defaultDateFormat);
```

This also keeps the DATE direction consistent with the string converter below, which already falls back when the format is empty.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeStringConverter.java
```diff
@@ -0,0 +1,81 @@
...
+            ZonedDateTime value, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        return new WriteCellData<>(value.format(formatter(contentProperty, globalConfiguration.getLocale())));
+    }
+
+    private DateTimeFormatter formatter(ExcelContentProperty contentProperty, Locale locale) {
+        if (contentProperty == null
+                || contentProperty.getDateTimeFormatProperty() == null
+                || StringUtils.isEmpty(
+                        contentProperty.getDateTimeFormatProperty().getFormat())) {
+            return DateTimeFormatter.ISO_ZONED_DATE_TIME;
```
Mikkey-f, line 77: `DateTimeFormatter.ofPattern(...)` is invoked for every cell conversion — once in `convertToJavaData` and once in `convertToExcelData`. In the per-cell hot path (large sheets) this rebuilds the same formatter over and over, since the pattern and locale change only per property, not per cell.

Suggest caching formatters per (pattern, locale), e.g. a `ThreadLocal` map with the ISO default kept as a shared constant:

```java
private static final ThreadLocal<Map<String, DateTimeFormatter>> FORMATTER_CACHE = new ThreadLocal<>();
```

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeStringConverter.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        return DateTimeFormatter.ofPattern(
+                contentProperty.getDateTimeFormatProperty().getFormat(), locale);
+    }
+}
```
Mikkey-f, line 82: Empty `@DateTimeFormat` values are handled inconsistently across the three converters:
- `ZonedDateTimeStringConverter` (line 81) falls back to `ISO_ZONED_DATE_TIME` → `2020-01-02T03:04:05Z`
- `ZonedDateTimeDateConverter` (line 54) passes `""` through to `fillDataFormat` → cell ends up with no format at all
- `ZonedDateTimeNumberConverter` → the cell is formatted with the workbook default

So the same `@DateTimeFormat(value = "")` produces three different outputs depending on the cell type. Suggest picking one behavior — falling back to the default `yyyy-MM-dd HH:mm:ss` on all three keeps it consistent with the other date-time families.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeDateConverter.java
alexsmolya: Fixed in f882ea6: empty DATE formats are normalized to null before fillDataFormat, so the existing yyyy-MM-dd HH:mm:ss default is applied. Added a regression test.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeStringConverter.java
alexsmolya: Addressed in f882ea6 by reusing the existing bounded per-thread locale/pattern formatter cache in DateUtils for configured patterns. No new converter-local ThreadLocal was introduced; ISO remains the shared constant.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeStringConverter.java
alexsmolya: The DATE defect is fixed in f882ea6. I left the three representations defaults intentionally distinct: STRING uses ISO_ZONED_DATE_TIME to preserve zone and offset text, while DATE and NUMBER use Excel date semantics and the workbook default. This matches the existing converter-family conventions, so a blanket yyyy-MM-dd HH:mm:ss fallback would change intended STRING behavior.

Mikkey-f:
Thanks for addressing all three points — verified locally, all tests green (8/8). One heads-up: the PR is currently in a CONFLICTING state (`DefaultConverterLoader`, after #1032 was merged to main) — could you rebase on the latest main when you get a chance?

nkuprins:
Please resolve the conflicts from main and fix the licence header as suggested by bengbengbalabalabeng :)

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/DateUtils.java
```diff
@@ -373,7 +373,14 @@ public static String format(BigDecimal date, Boolean use1904windowing, String da
...
 
-    private static DateTimeFormatter getCacheDateTimeFormat(String dateFormat, Locale locale) {
+    /**
+     * Get a cached date-time formatter for the supplied pattern and locale.
+     *
+     * @param dateFormat date-time pattern
+     * @param locale locale used to resolve the pattern
+     * @return cached formatter
+     */
+    public static DateTimeFormatter getCacheDateTimeFormat(String dateFormat, Locale locale) {
```
nkuprins, line 383: Making `getCacheDateTimeFormat` public exposes an internal cache accessor as API. The convention is to add a purpose-named helper to `DateUtils` instead:
- `DateUtils#parseZonedDateTime(String, String, Locale)` 
- `DateUtils#format(ZonedDateTime, String, Locale)`

It would also give `DateUtilsTest` something to test directly.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/zoneddatetime/ZonedDateTimeNumberConverter.java
```diff
@@ -0,0 +1,67 @@
...
+    public CellDataTypeEnum supportExcelTypeKey() {
+        return CellDataTypeEnum.NUMBER;
+    }
+
+    @Override
+    public ZonedDateTime convertToJavaData(
+            ReadCellData<?> cellData, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        boolean use1904windowing = globalConfiguration.getUse1904windowing();
+        if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
+            use1904windowing = contentProperty.getDateTimeFormatProperty().getUse1904windowing();
```
nkuprins: I suggest using the new method `DateUtils.isDate1904(contentProperty, globalConfiguration)`

Same advice for `convertToExcelData`.

fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/ZonedDateTimeConverterTest.java
```diff
@@ -0,0 +1,184 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fesod.sheet.converter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
```
nkuprins: Avoid static imports :)

PR #1022, opened 2026-08-21
Update dependabot configuration for GitHub Actions

pjfanning:
Removed ignore rules for major and minor updates in GitHub Actions.

It is against ASF recommendations to use old GitHub Actions. Supply Chain Attacks are a real issue.

I would question why you are blocking updates for other categories too and why you are not trying to get npm and jars up to date too. We have reached peak CVE with all the AI driven security alerts.


delei (approved):
LGTM

PR #1031, opened 2026-08-22
refactor: reuse Head#field in ModelBuildEventListener field-assignment fallback

bengbengbalabalabeng:

As title.

## What's changed?

- Replace the per-row `getDeclaredField(fieldName)` lookup with the `Field` already resolved on `Head`; the old lookup also missed superclass fields and silently dropped their values.
- Wrap `IllegalAccessException` in `ExcelDataConvertException` with row/column context instead of a bare `RuntimeException`.

## Checklist

### DIFF
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/read/listener/ModelBuildEventListenerTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/read/listener/ModelBuildEventListenerTest.java
@@ -24,6 +24,8 @@ import java.util.List;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.Accessors;
 import org.apache.fesod.sheet.annotation.ExcelProperty;
 import org.apache.fesod.sheet.testkit.Tags;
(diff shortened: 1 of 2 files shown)


fesod-sheet/src/test/java/org/apache/fesod/sheet/read/listener/ModelBuildEventListenerTest.java
```diff
@@ -0,0 +1,95 @@
...
+        @ExcelProperty("name")
+        public String name;
+    }
+
+    public static class InheritedGetterOnlyData extends GetterOnlyData {}
+
+    @Getter
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class PrivateGetterOnlyData {
```
delei, line 57: I re-read and retested the scenario in this old [issue](https://github.com/alibaba/easyexcel/issues/3524), and all of them passed the tests. Could you please add these test cases?


Non-standard set method

```Java

// Non-standard set method
public String setName(String name) {
    return this.name = name;
}
```


Use `@Accessors(chain = true)`

```Java
@Getter
@NoArgsConstructor
@AllArgsConstructor
@Accessors(chain = true)
public static class PrivateGetterOnlyData {
    ....
}
```

If I have missed anything, please feel free to add it.

delei (approved):
LGTM

PR #1032, opened 2026-08-23
feat: add java.time.LocalTime converters

leehaut:

Related: #1017

Add first-class converter support for `java.time.LocalTime`, following the existing `LocalDate` / `LocalDateTime` pattern.

## What's changed?

Excel has no native time-only cell type. This PR treats `LocalTime` as the time-of-day counterpart of `LocalDate`:

- Writing preserves the wall-clock time, attaches `DateUtils.EPOCH` (`1970-01-01`) as the date component, and applies `HH:mm:ss` so Excel shows time only.
- Reading drops any date component with `toLocalTime()`. An Excel-native fraction (`0.5` = `12:00:00`) and a full datetime serial that happens to be `12:00:00` both become `LocalTime.of(12, 0, 0)`.

Converter family under `org.apache.fesod.sheet.converters.localtime`:

- `LocalTimeDateConverter` — default write path (`DATE` cell, format `HH:mm:ss`)
- `LocalTimeNumberConverter` — bidirectional Excel numeric serial, including `use1904windowing`
- `LocalTimeStringConverter` — bidirectional `STRING` cell; default `HH:mm:ss`, also auto-detects `HH:mm`, respects `@DateTimeFormat` and `Locale`

`DateUtils` additions (mirroring `parseLocalDate` / `format(LocalDate)`):

- `DEFAULT_LOCAL_TIME_FORMAT` / `TIME_FORMAT_8` / `TIME_FORMAT_5`
- `parseLocalTime(...)` / `format(LocalTime, ...)` / `switchTimeFormat(...)` / `getLocalTime(...)`

Registration in `DefaultConverterLoader` matches `LocalDateTime`:

- Number and String converters in `initAllConverter()`
- Date converter as the default write converter
- String converter for string-formatted writes

Scope is limited to `java.time.LocalTime` (JDK 8). Converter lookup is keyed by the declared Java class, so existing `Date` converters do not cover `LocalTime` fields.

### DIFF
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/converters/localtime/LocalTimeStringConverterTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/converters/localtime/LocalTimeStringConverterTest.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fesod.sheet.converters.localtime;
+
+import java.time.LocalTime;
+import java.time.format.DateTimeParseException;
+import java.util.Locale;
+import java.util.stream.Stream;
+import org.apache.fesod.sheet.enums.CellDataTypeEnum;
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.ReadCellData;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.DateTimeFormatProperty;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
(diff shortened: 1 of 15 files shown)


bengbengbalabalabeng:

leehaut:
> The newly added files in this PR are implemented from scratch and are not derived from Alibaba's EasyExcel. Therefore, no EasyExcel-related license header is required for these files.
> 
> Please refer to: https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/FesodSheet.java

Thanks for the catch. Removed the EasyExcel-derived license header from the newly added LocalTime converter files; they now keep the Apache 2.0 header only, consistent with `FesodSheet.java`.

fesod-sheet/src/test/java/org/apache/fesod/sheet/converters/localtime/LocalTimeStringConverterTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        WriteCellData<?> written =
+                converter.convertToExcelData(LocalTime.of(12, 30, 45), contentProperty, GLOBAL_CONFIGURATION);
+
+        Assertions.assertEquals("12:30:45", written.getStringValue());
+    }
+
+    @ParameterizedTest
+    @MethodSource("locales")
+    void convertToExcelDataKeepsNumericTimeAcrossLocales(Locale locale) {
```
bengbengbalabalabeng, line 132: The `HH:mm:ss` pattern contains no locale-sensitive symbols, so its output is consistent across different Locales and cannot reflect conversion differences. If you want to test locale behavior, you can introduce the locale-sensitive field `a` (the AM/PM marker) and switch to `hh:mm:ss a`:

- Locale.US → 12:30:45 PM
- Locale.CHINA → 12:30:45 下午

fesod-sheet/src/test/java/org/apache/fesod/sheet/util/DateUtilsTest.java
```diff
@@ -19,12 +19,15 @@
 
 package org.apache.fesod.sheet.util;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
bengbengbalabalabeng: Avoid using wildcard imports.

bengbengbalabalabeng (approved):
LGTM.

PR #1033, opened 2026-08-23
feat: add java.time.OffsetDateTime converters (#1017)

Mikkey-f:
### What and why

Adds a `java.time.OffsetDateTime` converter family for the OffsetDateTime slice of #1017, following the existing `LocalDateTime` / `ZonedDateTime` pattern, so OffsetDateTime fields map to Excel natively instead of falling back to `String`:

- **`OffsetDateTimeDateConverter`** — write-only, emits an Excel `DATE` cell via `toLocalDateTime()`, default format `yyyy-MM-dd HH:mm:ss`
- **`OffsetDateTimeNumberConverter`** — bidirectional `NUMBER` serial, respects `use1904windowing` (property-level first, then a null-safe global default); on read attaches `ZoneId.systemDefault()` to the parsed `LocalDateTime`
- **`OffsetDateTimeStringConverter`** — bidirectional `STRING`, honors `@DateTimeFormat` and the configured `Locale`, defaults to `ISO_OFFSET_DATE_TIME` when no format is set; reads parse strictly and reject offset-less text (fail fast instead of a silent `ZoneId.systemDefault()` interpretation)

Registered in `DefaultConverterLoader.initAllConverter()` / `initDefaultWriteConverter()`.

### Tests

`OffsetDateTimeConverterTest` covers converter keys, DATE/NUMBER/STRING read & write, `@DateTimeFormat` formatting, `use1904windowing` (including the null-safe global default) and round-trip behavior. All tests pass, `spotless:check` is green, and the full local build was verified.

---

Related: #1017 (OffsetDateTime slice).

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * This file is part of the Apache Fesod (Incubating) project, which was derived from Alibaba EasyExcel.
+ *
+ * Copyright (C) 2018-2024 Alibaba Group Holding Ltd.
+ */
+
+package org.apache.fesod.sheet.converters.offsetdatetime;
+
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import org.apache.fesod.sheet.converters.Converter;
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+import org.apache.fesod.sheet.util.DateUtils;
+import org.apache.fesod.sheet.util.WorkBookUtil;
+
+/** OffsetDateTime and date converter. */
+public class OffsetDateTimeDateConverter implements Converter<OffsetDateTime> {
+    @Override
+    public Class<?> supportJavaTypeKey() {
+        return OffsetDateTime.class;
+    }
+
+    @Override
+    public WriteCellData<?> convertToExcelData(
+            OffsetDateTime value, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration)
+            throws Exception {
+        LocalDateTime localDateTime = value.toLocalDateTime();
+        WriteCellData<?> cellData = new WriteCellData<>(localDateTime);
+        String format = null;
+        if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
+            format = contentProperty.getDateTimeFormatProperty().getFormat();
+        }
+        WorkBookUtil.fillDataFormat(cellData, format, DateUtils.defaultDateFormat);
+        return cellData;
+    }
+}
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
@@ -27,6 +27,7 @@ package org.apache.fesod.sheet.converters.offsetdatetime;
 
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import org.apache.fesod.common.util.StringUtils;
 import org.apache.fesod.sheet.converters.Converter;
 import org.apache.fesod.sheet.metadata.GlobalConfiguration;
 import org.apache.fesod.sheet.metadata.data.WriteCellData;
@@ -50,6 +51,9 @@ public class OffsetDateTimeDateConverter implements Converter<OffsetDateTime> {
         String format = null;
         if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
             format = contentProperty.getDateTimeFormatProperty().getFormat();
+            if (StringUtils.isEmpty(format)) {
+                format = null;
+            }
         }
         WorkBookUtil.fillDataFormat(cellData, format, DateUtils.defaultDateFormat);
         return cellData;
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
@@ -17,12 +17,6 @@
  * under the License.
  */
 
-/*
- * This file is part of the Apache Fesod (Incubating) project, which was derived from Alibaba EasyExcel.
- *
- * Copyright (C) 2018-2024 Alibaba Group Holding Ltd.
- */
-
 package org.apache.fesod.sheet.converters.offsetdatetime;
 
 import java.time.LocalDateTime;
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeNumberConverter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeNumberConverter.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * This file is part of the Apache Fesod (Incubating) project, which was derived from Alibaba EasyExcel.
+ *
+ * Copyright (C) 2018-2024 Alibaba Group Holding Ltd.
+ */
+
+package org.apache.fesod.sheet.converters.offsetdatetime;
+
+import java.math.BigDecimal;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import org.apache.fesod.sheet.converters.Converter;
+import org.apache.fesod.sheet.enums.CellDataTypeEnum;
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.ReadCellData;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+import org.apache.fesod.sheet.util.DateUtils;
+import org.apache.poi.ss.usermodel.DateUtil;
+
+/** OffsetDateTime and number converter. */
+public class OffsetDateTimeNumberConverter implements Converter<OffsetDateTime> {
+    @Override
+    public Class<?> supportJavaTypeKey() {
+        return OffsetDateTime.class;
+    }
+
+    @Override
+    public CellDataTypeEnum supportExcelTypeKey() {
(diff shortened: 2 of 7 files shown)


fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeDateConverter.java
```diff
@@ -0,0 +1,57 @@
...
+    public WriteCellData<?> convertToExcelData(
+            OffsetDateTime value, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration)
+            throws Exception {
+        LocalDateTime localDateTime = value.toLocalDateTime();
+        WriteCellData<?> cellData = new WriteCellData<>(localDateTime);
+        String format = null;
+        if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
+            format = contentProperty.getDateTimeFormatProperty().getFormat();
+        }
+        WorkBookUtil.fillDataFormat(cellData, format, DateUtils.defaultDateFormat);
```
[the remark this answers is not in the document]
Mikkey-f: Empty @DateTimeFormat format — Fixed in OffsetDateTimeDateConverter: empty formats are now normalized to null before calling WorkBookUtil.fillDataFormat, so the yyyy-MM-dd HH:mm:ss default is applied. Added a regression test(dateConverterFallsBackToDefaultFormatForEmptyDateTimeFormat).

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeStringConverter.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private DateTimeFormatter formatter(ExcelContentProperty contentProperty, Locale locale) {
+        if (locale == null) {
+            locale = Locale.getDefault();
+        }
+        String format = format(contentProperty);
+        if (StringUtils.isEmpty(format)) {
+            return DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+        }
+        return DateTimeFormatter.ofPattern(format, locale);
```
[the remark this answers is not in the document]
Mikkey-f: Formatter caching — Fixed in OffsetDateTimeStringConverter: DateTimeFormatter instances are now cached per pattern and locale in a thread-local map, avoiding rebuilds in the per-cell hot path (the ISO default is a shared constant).

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeNumberConverter.java
```diff
@@ -0,0 +1,83 @@
...
+                value.toLocalDateTime(), resolveUse1904windowing(contentProperty, globalConfiguration))));
+    }
+
+    private boolean resolveUse1904windowing(
+            ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        if (contentProperty != null && contentProperty.getDateTimeFormatProperty() != null) {
+            Boolean propertyUse1904windowing =
+                    contentProperty.getDateTimeFormatProperty().getUse1904windowing();
+            if (propertyUse1904windowing != null) {
+                return propertyUse1904windowing;
```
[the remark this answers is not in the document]
Mikkey-f: use1904windowing DEFAULT being unboxed to false — Agreed this is a real issue, but it's a pre-existing framework-level problem: DateTimeFormatProperty.build converts BooleanEnum.DEFAULT (null) to false (DateTimeFormatProperty.java:55-57), and all existing number converters (Date, LocalDate, LocalDateTime, ZonedDateTime) consume the property value without a global fallback. This PR's null-safe fallback covers the no-annotation path; the annotated path behaves identically to the existing converter families. Fixing it properly means changing DateTimeFormatProperty (preserving DEFAULT as null) and updating every date-number converter — a framework-wide change that deserves its own issue/PR. Happy to open one if that's useful.

bengbengbalabalabeng:

fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/OffsetDateTimeConverterTest.java
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * Copyright (C) 2018-2024 Alibaba Group Holding Ltd.
+ */
+
+package org.apache.fesod.sheet.converter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
bengbengbalabalabeng: Avoid using wildcard imports.

bengbengbalabalabeng:
> The newly added files in this PR are implemented from scratch and are not derived from Alibaba's EasyExcel. Therefore, no EasyExcel-related license header is required for these files.
> 
> Please refer to: https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/FesodSheet.java

Please update the license header.

Mikkey-f:
Thanks for the review. You're right — these files were implemented from scratch, so I've removed the EasyExcel-derived header block (including the Alibaba copyright notice) from the three new converters and the test file. They now carry only the standard ASF header, matching the convention in FesodSheet.java and the LocalTime converters merged in #1032.

fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/OffsetDateTimeConverterTest.java
Mikkey-f: Thanks for the comment. The imports this is anchored to (lines 22-25, after the header cleanup: `assertEquals`, `assertNotNull`, `assertNull`, `assertThrows`) are explicit static imports — each imports a single method, and all four are used in the tests. There is no `import static org.junit.jupiter.api.Assertions.*;` style wildcard anywhere in the files changed by this PR.

If you mean these four lines should not be collapsed into a wildcard import — current code already follows that. If you'd rather I drop the static imports entirely and use fully-qualified `Assertions.xxx` calls, I'm happy to do that — just confirm and I'll update.

bengbengbalabalabeng: Apologies for the typo. The intended phrase was "avoid static imports".

Mikkey-f: Thanks for the clarification — I've replaced all static Assertions imports in OffsetDateTimeConverterTest with fully-qualified `Assertions.xxx` calls (16 call sites, plus the import). The other files in this PR had no static imports. Updated and pushed.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeStringConverter.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+import org.apache.fesod.sheet.util.DateUtils;
+
+/** OffsetDateTime and string converter. */
+public class OffsetDateTimeStringConverter implements Converter<OffsetDateTime> {
+    /**
+     * Thread-local cache of {@link DateTimeFormatter} instances, keyed by pattern and locale, so
+     * the per-cell hot path does not rebuild formatters for every conversion.
+     */
+    private static final ThreadLocal<Map<String, DateTimeFormatter>> FORMATTER_CACHE = new ThreadLocal<>();
```
bengbengbalabalabeng: Since `DateTimeFormatter` caching logic is already implemented in `DateUtils` via `DATE_TIME_FORMATTER_THREAD_LOCAL` (which has a proper cleanup lifecycle), could we consolidate this into `DateUtils` (e.g., `DateUtils#parseOffsetDateTime`, `#format(OffsetDateTime...)`) instead of maintaining an isolated `ThreadLocal` here?

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeStringConverter.java
```diff
@@ -0,0 +1,112 @@
...
+                        .toOffsetDateTime();
+            } catch (RuntimeException inner) {
+                if (StringUtils.isEmpty(format)) {
+                    return LocalDateTime.parse(stringValue, DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+                            .atZone(ZoneId.systemDefault())
+                            .toOffsetDateTime();
+                }
+                throw inner;
+            }
+        }
```
bengbengbalabalabeng: Silently falling back to `ZoneId.systemDefault()` can lead to hidden bugs, the exact same non-offset string would resolve to completely different points in time depending on the server's timezone (e.g., local UTC+8 vs. UTC in production).

I suggest removing it in favor of a fail-fast approach: Parse strictly using the specified pattern if provided, or default to `yyyy-MM-dd'T'HH:mm:ssXXX` if absent. If parsing fails, let it throw directly so the caller knows the data does not match the expected format.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeStringConverter.java
Mikkey-f: Agreed — I've consolidated this into `DateUtils`. It now provides `DateUtils#parseOffsetDateTime(String, String, Locale)` and a `format(OffsetDateTime, String, Locale)` overload, both backed by the bounded `DATE_TIME_FORMATTER_THREAD_LOCAL` cache that `removeThreadLocalCache()` clears at the end of each read/write context. `OffsetDateTimeStringConverter` now delegates to these helpers (as `LocalDateTimeStringConverter` does) and no longer keeps its own converter-local `ThreadLocal`. When the format is empty, both helpers fall back to `ISO_OFFSET_DATE_TIME`.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/offsetdatetime/OffsetDateTimeStringConverter.java
Mikkey-f: Agreed — offset-less text should not be silently re-interpreted in `ZoneId.systemDefault()`, since the same text would then resolve to different instants depending on the server timezone. I've removed the fallback: STRING reads now parse strictly with the configured pattern, or `ISO_OFFSET_DATE_TIME` when no format is set, and propagate `DateTimeParseException` for text without an explicit offset. Note this also applies to configured patterns without an offset section, which can only ever produce offset-less text — wall-clock-only columns are better served by the `LocalDateTime` converters. Tests updated: the two previous fallback cases now assert failure, and a round-trip case for a configured offset pattern was added.

bengbengbalabalabeng (dismissed):
LGTM.

nkuprins:
The tests don't seem to pin the locale? If you swap it for a different locale, nothing fails. Maybe use something locale-sensitive like _"dd MMMM yyyy HH:mm:ss XXX"_?

nkuprins:
2 new methods were added to `DateUtils` by your last refactoring. However, no new tests for them were added to `DateUtilsTest`? Maybe we can add something like this to `DateUtilsTest` to keep it consistent:

```java
@Test
void test_parseOffsetDateTime() {
    OffsetDateTime expected = OffsetDateTime.of(2020, 1, 2, 3, 4, 5, 0, ZoneOffset.ofHours(8));
    Assertions.assertEquals(expected, DateUtils.parseOffsetDateTime("2020-01-02T03:04:05+08:00", null, Locale.US));
    Assertions.assertEquals(expected, DateUtils.parseOffsetDateTime("2020-01-02T03:04:05+08:00", "", Locale.US));
    Assertions.assertEquals(
            expected,
            DateUtils.parseOffsetDateTime("02 Januar 2020 03:04:05 +08:00", "dd MMMM yyyy HH:mm:ss XXX", Locale.GERMAN));
    Assertions.assertThrows(
            DateTimeParseException.class,
            () -> DateUtils.parseOffsetDateTime("2020-01-02T03:04:05", null, Locale.US));
}

@Test
void test_format_OffsetDateTime() {
    OffsetDateTime value = OffsetDateTime.of(2020, 1, 2, 3, 4, 5, 0, ZoneOffset.ofHours(8));
    Assertions.assertNull(DateUtils.format((OffsetDateTime) null, null, Locale.US));
    Assertions.assertEquals("2020-01-02T03:04:05+08:00", DateUtils.format(value, null, Locale.US));
    Assertions.assertEquals("2020-01-02T03:04:05+08:00", DateUtils.format(value, "", Locale.US));
    Assertions.assertEquals(
            "02 Januar 2020 03:04:05 +08:00", DateUtils.format(value, "dd MMMM yyyy HH:mm:ss XXX", Locale.GERMAN));
}
```

nkuprins:
Sibling test classes to `OffsetDateTimeConverterTest` seem to have:
```java
@AfterEach
void tearDown() {
    DateUtils.removeThreadLocalCache();
}
```
Should we also add it for consistency?

Mikkey-f:
Thanks for the review — all three are addressed in 7653c542 (test-only, no production code changed):

1. Added `stringConverterHonoursConfiguredLocale`, which pins `Locale.GERMAN` with a locale-sensitive pattern (`dd MMMM yyyy HH:mm:ss XXX`) and asserts both the written text and the round-trip read, so the locale is now actually exercised and independent of the machine default.
2. Added `test_parseOffsetDateTime` and `test_format_OffsetDateTime` to `DateUtilsTest`, following the cases you outlined.
3. Added the `@AfterEach tearDown()` calling `DateUtils.removeThreadLocalCache()`, matching the sibling converter tests.

spotless and the full test suite (932 tests) are green.

Mikkey-f:
@bengbengbalabalabeng a quick heads-up on the latest push: 7653c542 only adds tests, covering the three points nkuprins raised — locale-sensitive coverage for the string converter, direct tests for the two new `DateUtils` helpers, and the `@AfterEach` thread-local cache cleanup to match the sibling tests. No production code changed; the converters you approved are untouched.

The new commit auto-dismissed your approval, so it needs a fresh approve before it can be merged. Sorry for the extra round trip.

bengbengbalabalabeng (approved):
LGTM.

PR #1034, opened 2026-08-23
fix: resolve Windows fill template copy failures (#1021)

### LINKED ISSUE #1021: [Bug] Fill tests fail on Windows: output stream leak on CSV+template exception, and URL.getPath() breaks example templates
### Search before asking

- [x] I searched in the issues and found nothing similar.

### Fesod version

main (latest)

### JDK version

17 (Temurin)

### Operating system

Windows 11

### Steps To Reproduce

Run `mvn test -Dmaven.test.skip=false` and `mvn verify` on Windows:
- `fesod-sheet` module: `FillDataTest.simpleFill(ExcelFormat)[3]` (CSV variant) fails — JUnit reports `Failed to close extension context`: the temp CSV file cannot be deleted because it is still locked
- `fesod-examples/fesod-sheet-examples`: `FillBasicExampleITCase` and `FillComplexExampleITCase` fail with `ExcelGenerateException: Copy template failure`

CI (Ubuntu) is green, so these are Windows-specific failures.

### Current Behavior

**1. Output stream leak on the CSV+template exception path (product bug, not limited to Windows):**
`WriteWorkbookHolder`'s constructor opens the output stream first (`new FileOutputStream(file)`, [WriteWorkbookHolder.java#L197](https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java#L197)) and then calls `copyTemplate()`, which throws `ExcelGenerateException("csv cannot use template.")` for CSV + template combinations ([WriteWorkbookHolder.java#L278](https://github.com/apache/fesod/blob/main/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java#L278)). The exception aborts the constructor and the stream is never closed. The file handle is leaked on every platform; on Windows the file stays locked, which surfaces as the `FillDataTest` failure when JUnit's @TempDir cleanup cannot delete the file.

**2. `URL.getPath()` returns a URL-encoded path on Windows (example infrastructure bug):**
`ExampleFileUtil.getExamplePath(...)` resolves classpath resources via `getResource(...).getPath()`. On Windows the resource URL encodes the path separator as `%5c` (e.g. `.../example/templates%5csimple.xlsx`), so the returned path is never decoded and `new File(...)` throws `FileNotFoundException`. Example fill templates cannot be loaded on Windows at all.

### Expected Behavior

- The output stream should be closed when the constructor aborts (exception-safe resource handling), so no file handle is leaked and Windows tests pass.
- `ExampleFileUtil` should return a properly decoded path (e.g. via `URL.toURI()`), so examples work on Windows.

### Anything else?

Both are reproducible on a clean checkout on Windows 11. The stream leak in (1) is a cross-platform resource leak that CI currently masks.

Mikkey-f:
### What and why

Fixes the `Copy template failure` errors that the fill examples
(`FillBasicExample`, `FillComplexExample`) hit on Windows. Two root causes:

**1. `WriteWorkbookHolder` leaks the output stream it opens when initialization fails**

The constructor opens a `FileOutputStream` for the target file before copying
the template. When the copy fails (or any later initialization step throws),
the stream was never closed — the file stays locked on Windows. Unix lets you
delete an open file, which is why CI stayed green. This is a cross-platform
resource leak; Windows just surfaces it first.

Fix: wrap `initHandler` + `copyTemplate` so the stream opened by the holder is
closed before the exception propagates. Caller-provided streams are untouched —
ownership stays with the caller.

**2. `ExampleFileUtil` returns the percent-encoded URL path**

`URL.getPath()` does not decode. On Windows, template file names built with
`File.separator` (`\`) are encoded as `%5C` in the resource URL, so the returned
path pointed at a file literally named `templates%5Clist.xlsx` — which does not
exist → `Copy template failure`. On Unix the separator `/` needs no encoding,
which is why CI passed.

Fix: decode via `URL.toURI()` in a shared `toFilePath` helper. This also fixes
paths containing spaces or other reserved characters (`%20` etc.).

### Tests

- New `WriteWorkbookHolderOutputStreamTest` (2 cases): the stream opened by the
  holder is closed on init failure (on Windows an open stream locks the file, so
  deletion fails without the fix); caller-provided streams are not closed.
- The fill ITCases (`FillBasicExampleITCase`, `FillComplexExampleITCase`) now
  pass on Windows; they previously failed with `Copy template failure`.
- Full module suite green: 846 tests, 0 failures (fesod-common / fesod-shaded /
  fesod-sheet / fesod-sheet-examples), `spotless:check` and `javadoc:javadoc` pass.

---

Closes #1021

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java
@@ -244,10 +244,10 @@ public class WriteWorkbookHolder extends AbstractWriteHolder {
             initHandler(writeWorkbook, null);
             copyTemplate();
         } catch (IOException e) {
-            closeOutputStreamIfOwned();
+            closeOutputStreamOnFailure();
             throw new ExcelGenerateException("Copy template failure.", e);
         } catch (RuntimeException e) {
-            closeOutputStreamIfOwned();
+            closeOutputStreamOnFailure();
             throw e;
         }
         if (writeWorkbook.getMandatoryUseInputStream() == null) {
@@ -274,12 +274,13 @@ public class WriteWorkbookHolder extends AbstractWriteHolder {
     }
 
     /**
-     * Close the output stream opened by this holder when initialization fails, so that the target file
-     * is not held open. Only streams opened from {@code file} are closed; streams passed in by the
-     * caller remain the caller's responsibility.
+     * Close the output stream when initialization fails, so that the target file is not held open.
+     * Mirrors the close behavior of the success path, which is also gated by {@code autoCloseStream};
+     * callers that opt out of automatic closing via {@code autoCloseStream(false)} keep that behavior
+     * here as well.
      */
-    private void closeOutputStreamIfOwned() {
-        if (file != null && outputStream != null) {
+    private void closeOutputStreamOnFailure() {
+        if (autoCloseStream && outputStream != null) {
             try {
                 outputStream.close();
             } catch (IOException e) {
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java
@@ -240,12 +240,15 @@ public class WriteWorkbookHolder extends AbstractWriteHolder {
         }
 
         // init handler
-        initHandler(writeWorkbook, null);
-
         try {
+            initHandler(writeWorkbook, null);
             copyTemplate();
         } catch (IOException e) {
+            closeOutputStreamIfOwned();
             throw new ExcelGenerateException("Copy template failure.", e);
+        } catch (RuntimeException e) {
+            closeOutputStreamIfOwned();
+            throw e;
         }
         if (writeWorkbook.getMandatoryUseInputStream() == null) {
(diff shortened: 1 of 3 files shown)


fesod-sheet/src/main/java/org/apache/fesod/sheet/write/metadata/holder/WriteWorkbookHolder.java
```diff
@@ -240,12 +240,15 @@ public WriteWorkbookHolder(WriteWorkbook writeWorkbook) {
         }
 
         // init handler
-        initHandler(writeWorkbook, null);
-
         try {
+            initHandler(writeWorkbook, null);
             copyTemplate();
         } catch (IOException e) {
+            closeOutputStreamIfOwned();
```
delei: Could we gate `closeOutputStreamIfOwned()` with `autoCloseStream` instead of `file != null?`

`autoCloseStream` is more consistent with existing close behavior (while still not closing caller-provided streams).

Mikkey-f: Thanks for the suggestion — updated. The failure path in `WriteWorkbookHolder` now gates the close on `autoCloseStream` instead of `file != null`, which mirrors the success-path behavior. Callers using `autoCloseStream(false)` keep their streams unclosed on failure as well, and the test now sets `autoCloseStream(false)` explicitly to assert that contract.

delei (approved):
LGTM

PR #1036, opened 2026-08-24
fix: throw exception when headClazz has no valid fields

bengbengbalabalabeng:
## What's changed?

Added validation in `initColumnProperties` to throw an exception when `headClazz` has no valid fields (fail-fast).

## Checklist


bengbengbalabalabeng:
Given the scenario in https://github.com/apache/fesod/blob/main/fesod-sheet/src/test/java/org/apache/fesod/sheet/core/ClassUtilsFieldOverrideTest.java, fail-fast may not be appropriate.

Close for now.

PR #1054, opened 2026-08-28
fix: fall back to global 1904 windowing when @DateTimeFormat default is used (#1042)

### LINKED ISSUE #1042: [Bug] @DateTimeFormat use1904windowing DEFAULT is collapsed to false, bypassing the global 1904 fallback
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar. (#433/#434 cover the global read/write 1904 support and are unrelated to the annotation's `DEFAULT` semantics.)

### Fesod version

2.0.2-incubating+ (current main)

### JDK version

JDK 8+ (reproducible on any)

### Operating system

Any

### Steps To Reproduce

1. Create a POJO with a `LocalDateTime` (or `Date`) field annotated with `@DateTimeFormat` for the pattern only, leaving `use1904windowing` at its default:

```java
public class DemoData {
    @ExcelProperty("time")
    @DateTimeFormat("yyyy-MM-dd HH:mm:ss")   // use1904windowing left at DEFAULT
    private LocalDateTime time;
}
```

2. Read a workbook whose date windowing is 1904, either by builder setting or by a file carrying the 1904 flag:

```java
FesodSheet.read(path)
    .head(DemoData.class)
    .use1904windowing(true)   // global setting
    .sheet()
    .doReadSync();
```

3. Observe the parsed `LocalDateTime` — serial numbers are interpreted with the 1900 window (offset by 1462 days) instead of 1904.

### Current Behavior

`DateTimeFormatProperty.build()` collapses the annotation's `use1904windowing()` default (`BooleanEnum.DEFAULT` = null) to `false`:

```java
return new DateTimeFormatProperty(
        dateTimeFormat.value(),
        BooleanUtils.isTrue(dateTimeFormat.use1904windowing().getBooleanValue()));
```

So whenever a field carries `@DateTimeFormat` (even for the pattern only), the date-number converters consume `false` and the global `use1904windowing` setting is bypassed:

| Path | 1904 value used |
|---|---|
| No annotation | `globalConfiguration.getUse1904windowing()` (e.g. `.use1904windowing(true)` on the builder) |
| `@DateTimeFormat` with `DEFAULT` | `false` — **global setting is bypassed** |
| `@DateTimeFormat(use1904windowing = TRUE/FALSE)` | explicit value (correct) |

The same POJO behaves differently on the read path (serial → date) and write path (date → serial) depending on whether the field happens to carry the annotation.

### Expected Behavior

`DEFAULT` should keep its "user did not specify" semantics (as `BooleanEnum`'s javadoc states — an enum "to determine whether the user has added the enumeration"). The converters should fall back to `globalConfiguration.getUse1904windowing()` when the property value is null, with the existing null-safe global default of `false` unchanged.

Suggested fix:
1. `DateTimeFormatProperty.build()`: pass `use1904windowing().getBooleanValue()` through as-is (keep `DEFAULT` as `null`)
2. Date-number converters (`Date`, `LocalDate`, `LocalDateTime`, `LocalTime`, `ZonedDateTime`, `OffsetDateTime`): when the property value is null, fall back to the global configuration — the `OffsetDateTimeNumberConverter` already implements this pattern
3. Regression test: `@DateTimeFormat` without `use1904windowing` + global `.use1904windowing(true)` → follows the global setting

### Anything else?

Historical note from #433: `GlobalConfiguration`'s `use1904windowing = Boolean.FALSE` default was introduced in 2021 to fix a test NPE — that null-safe global default stays untouched by this proposal.

Mikkey-f:
> **Summary**
>
> Fixes #1042. When a field carries `@DateTimeFormat` for the pattern only, its `use1904windowing` is left at `BooleanEnum.DEFAULT`; `DateTimeFormatProperty.build()` collapsed that to `false`, so the global 1904 windowing setting (builder option or workbook flag) was silently bypassed and date-number cells were interpreted with the 1900 windowing system, shifting dates by 1462 days (4 years).
>
> **Changes**
>
> - `DateTimeFormatProperty.build()`: preserve `DEFAULT` as `null` instead of collapsing it to `false`
> - New shared helper `DateUtils.resolveUse1904windowing(contentProperty, globalConfiguration)`: explicit annotation value wins, otherwise fall back to the global configuration (null-safe, defaults to `false`)
> - `DateNumberConverter`, `LocalDateNumberConverter`, `LocalDateTimeNumberConverter`, `LocalTimeNumberConverter`, `StringNumberConverter`: use the shared helper on both read and write paths
> - `TimestampNumberConverter` test fixture and the custom-converter example in `core-class.md` (EN/zh-CN) updated to the new helper, since they previously unboxed the property value directly
>
> **Testing**
>
> - New unit tests: `DateTimeFormatPropertyTest` (DEFAULT/TRUE/FALSE preserved), per-converter regression tests covering annotation-default + global 1904 → follows global; explicit annotation → annotation wins
> - New end-to-end regression test `DateTimeFormatDefaultWindowingTest` replicating the issue's reproduction steps: on the pre-fix code it fails with `expected: <2026-08-28T10:30> but was: <2022-08-27T10:30>` (the exact 4-year shift from the issue); on the fixed code it passes
> - Full suite: 913/913 green, spotless and javadoc pass
>
> **Note**
>
> Verified empirically that the write path already followed the global setting before this change (the write API consumes the annotation's `use1904windowing` only via the shared converters). The write-path test is kept as an invariant guard rather than a bug regression test. `OffsetDateTime`/`ZonedDateTime` converters on #1033 already implement the same fallback pattern and are unaffected.


fesod-sheet/src/main/java/org/apache/fesod/sheet/util/DateUtils.java
```diff
@@ -373,6 +375,27 @@ public static String format(BigDecimal date, Boolean use1904windowing, String da
...
 
+    /**
+     * Resolve the 1904 windowing flag for a field: the annotation value wins when explicitly set, otherwise fall back
+     * to the global configuration (null-safe, defaults to false).
+     *
+     * @param contentProperty the field's content property, may be null
+     * @param globalConfiguration the global configuration
+     * @return true if dates use the 1904 windowing system
+     */
+    public static boolean resolveUse1904windowing(
```
bengbengbalabalabeng: The current method name `resolveUse1904windowing` sounds like an action while it returns a boolean.

How about renaming it to `isDate1904`? Makes it clearer that it is a boolean method. (Reference: Apache POI - `DataFormatter#isDate1904(Cell)`)

fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/property/DateTimeFormatProperty.java
```diff
@@ -52,8 +51,9 @@ public static DateTimeFormatProperty build(DateTimeFormat dateTimeFormat) {
...
             return null;
         }
+        // Keep DEFAULT as null so converters can fall back to the global configuration
+        // instead of collapsing it to false (see #1042).
         return new DateTimeFormatProperty(
-                dateTimeFormat.value(),
-                BooleanUtils.isTrue(dateTimeFormat.use1904windowing().getBooleanValue()));
+                dateTimeFormat.value(), dateTimeFormat.use1904windowing().getBooleanValue());
     }
 }
```
bengbengbalabalabeng, line 65: Could we also add a Javadoc on the `use1904windowing` field explaining the behavior? Specifically, noting that `null` represents the `DEFAULT` state which indicates falling back to `GlobalConfiguration`, whereas `TRUE/FALSE` represents an explicit override.

This will prevent future maintainers from inadvertently collapsing `null` into `false` again.

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/DateUtils.java
Mikkey-f: Thanks for the suggestion — I've renamed it to `isDate1904` in cb67fb1. The Javadoc now states the resolution order explicitly (annotation override first, then fall back to the global configuration).

fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/property/DateTimeFormatProperty.java
Mikkey-f: Done in cb67fb1 — added the field Javadoc noting that `null` represents the `DEFAULT` state (fall back to the global configuration) while `TRUE`/`FALSE` are explicit overrides, so future maintainers don't collapse `null` into `false` again.

bengbengbalabalabeng (approved):
LGTM.

PR #1060, opened 2026-08-30
chore: remove unused mockserver-netty test dependency

leehaut:
## Summary
- Drop unused `org.mock-server:mockserver-netty` 5.15.0 from the parent POM (`mockserver.version`, `dependencyManagement`, and the inherited test dependency).
- The only usage was `WebStreamReadTest`, which mocked a remote `.xlsx` URL. That test and `web/io.xlsx` were removed in #770 when examples were refactored
- This also removes transitive BouncyCastle 1.72 findings: CVE-2024-29857, CVE-2024-30172, CVE-2024-34447, and CVE-2024-30171. MockServer 6+/7.x cannot be used here because it requires Java 17, while this project still tests on JDK 8.


delei (approved):
LGTM

PR #781, opened 2026-01-08
feature: Add the support module to meet ASF policy requirements (#780)

psxjoy:
## Purpose of the pull request
Add the support module to meet ASF policy requirements 

## Checklist

- [X] I have read the [Contributor Guide](https://fesod.apache.org/community/contribution/).
- [X] I have written the necessary doc or comment.
- [X] I have added the necessary unit tests and all cases have passed.

PR #1067, opened 2026-09-05
chore: update gitattributes to exclude non-source files

delei:

As title.

## What's changed?

Added `export-ignore` rules to exclude non-source artifacts from git archive.

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -26,7 +26,7 @@ mvnw.cmd text eol=crlf
 
 # Exclude non-source artifacts from `git archive`
 .idea/ export-ignore
-.mvn/maven-wrapper.jar export-ignore
+.mvn/wrapper/maven-wrapper.jar export-ignore
 .github/ export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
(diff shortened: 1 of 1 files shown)


alaahong (dismissed):
Thanks for the cleanup — this covers the `maven-wrapper.jar` that ended up in the RC3 source-release zip (raised in the [vote thread](https://lists.apache.org/thread/3mmsqnmxy7b6stllg8bkp08t8z6wc8z8)).

Two more tracked, non-source paths that could join the list, if we want the archive to contain only what's needed to build:

- `.asf.yaml` — ASF infra config for the GitHub repo, not needed when building from source
- `dist/` — currently only `dist/licenses/*.txt` (third-party license texts for the binary distribution). No build references anywhere, and `dist/**` is already in the root pom's apache-rat excludes, so ignoring it here keeps the two consistent

Keeping `mvnw`, `mvnw.cmd` and `.mvn/wrapper/maven-wrapper.properties` in the archive is correct — `mvnw` bootstraps the wrapper jar from the properties file.

Both additions are minor and optional — the PR is good as-is, just flagging them for completeness.

delei:
Referencing PR #781 , the license file in the `dist` directory is required by the ASF policies, I think it's still necessary to retain.

alaahong (approved):

alaahong:
LGTM

PR #1068, opened 2026-09-05
fix: make DataFormatter cache configuration-aware

### LINKED ISSUE #1065: [Bug] Thread-local DataFormatter reuses a stale locale across operations
### Search before asking

### Fesod version

2.1.0-incubating-rc3 (commit `1cd6e891316cfea2c49028091be9d6ee0409cd1e`)

### JDK version

Oracle JDK 11.0.17

### Operating system

Windows 11 (amd64), with the JVM default locale set to Korean

### Steps To Reproduce

I found this while checking the 2.1.0-incubating RC3 source release. I ran the full test suite twice from a clean source archive:

```powershell
.\mvnw.cmd clean verify "-Dmaven.test.skip=false" "-DskipTests=false"
```

Both runs failed at the same test in `fesod-sheet`:

```text
Tests run: 894, Failures: 1, Errors: 0, Skipped: 0
DateFormatTest.readCn(ExcelFormat)[1]
```

`DateFormatTest` passes when run by itself. I can reproduce the failure with this order on the same thread:

```text
CsvRowTest#csvWrite_withSqlDateAndTime_producesCorrectFile
DateFormatTest
```

I used this command for the ordered run:

```powershell
.\mvnw.cmd -pl fesod-sheet "-Dmaven.test.skip=false" "-DskipTests=false" '-Dtest=CsvRowTest#csvWrite_withSqlDateAndTime_producesCorrectFile,DateFormatTest' test
```

For a control, I temporarily added this call immediately after the CSV operation:

```java
NumberDataFormatterUtils.removeThreadLocalCache();
```

With that call in place, the same ordered run passed with 6 tests and 0 failures. I reverted the change after the run.

### Current Behavior

The Chinese date-format test gets a value containing the Korean AM marker (`오전`) instead of the expected Chinese marker (`上午`), even though the read operation sets `Locale.CHINA`.

`NumberDataFormatterUtils` stores one `DataFormatter` per thread:

```java
DataFormatter dataFormatter = DATA_FORMATTER_THREAD_LOCAL.get();
if (dataFormatter == null) {
    dataFormatter = new DataFormatter(use1904windowing, locale, useScientificFormat);
    DATA_FORMATTER_THREAD_LOCAL.set(dataFormatter);
}
```

After the first call, the cached formatter is reused without checking whether `locale`, `use1904windowing`, or `useScientificFormat` has changed.

In this reproducer, the direct `CsvWorkbook`/`CsvSheet` operation creates the formatter with the JVM's default Korean locale. Closing the sheet does not clear the `NumberDataFormatterUtils` cache. The following read requests `Locale.CHINA` on the same thread, but it receives the cached Korean formatter.

The result therefore depends on which operation ran first. Clearing the cache between the two operations removes the failure. I have not confirmed whether the direct low-level CSV API is expected to clear this cache, or whether `NumberDataFormatterUtils` should recreate its formatter when the requested configuration changes.

### Expected Behavior

A read configured with `Locale.CHINA` should use Chinese date markers regardless of an earlier operation's locale on the same thread.

The formatter should be recreated when its effective configuration changes, or each supported lifecycle that initializes it should clear the cache before another operation can reuse it.

### Anything else?

I reported the same failure in my RC3 vote on the Fesod development mailing list:

https://www.mail-archive.com/dev@fesod.apache.org/msg00235.html

There has not been a reply about the locale behavior yet, so I am also recording the reproducer here. The same `NumberDataFormatterUtils` cache logic is present on the current `main` branch (`b8adc73fd384155093c1769d6c53d312f9008cde`).

I found these related reports, but they cover different problems:

- #1051 covers sheet-level configuration not being copied by `ReadSheet.copyBasicParameter`.
- #1040 covers `java.util.Date` string conversion not receiving `GlobalConfiguration.locale`.
- #948 added locale-aware caches to `DateUtils` and `NumberUtils`, but did not change `NumberDataFormatterUtils`.

I am willing to submit a regression test and a fix. Before starting the PR, I would like to confirm whether the preferred fix is to make this cache configuration-aware or to clear it as part of the direct CSV lifecycle.

skytin1004:

Closed: #1065

`NumberDataFormatterUtils` currently reuses the first `DataFormatter` created on a thread even when a later operation requests a different locale, date windowing mode, or scientific-format setting. This makes formatting order-dependent and can produce output in the wrong locale.

## What's changed?

- Cache the `DataFormatter` together with its effective constructor configuration.
- Reuse the cached formatter only when `use1904windowing`, `locale`, and `useScientificFormat` all match; otherwise replace it.
- Add regression tests for sequential locale, 1904-windowing, and scientific-format changes while retaining coverage for cache reuse and explicit removal.
- Protect the test that changes the JVM default locale with JUnit's locale resource lock so it remains isolated under parallel test execution.

Verification:

- `fesod-sheet` test suite with the JVM default locale set to Korean: 897 tests passed.
- Locale-sensitive utility suites with the `parallel-tests` profile: 100 tests passed.
- GitHub Actions passed on JDK 8, 11, 17, 21, and 25, together with Spotless, license, and CodeQL checks.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/NumberDataFormatterUtils.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/NumberDataFormatterUtils.java
@@ -108,10 +108,8 @@ public class NumberDataFormatterUtils {
         Locale resolvedLocale = locale == null ? Locale.getDefault() : locale;
         boolean resolvedUseScientificFormat = Boolean.TRUE.equals(useScientificFormat);
         DataFormatterCache cache = DATA_FORMATTER_THREAD_LOCAL.get();
-        if (cache == null
-                || !cache.matches(resolvedUse1904windowing, resolvedLocale, resolvedUseScientificFormat)) {
-            cache = new DataFormatterCache(
-                    resolvedUse1904windowing, resolvedLocale, resolvedUseScientificFormat);
+        if (cache == null || !cache.matches(resolvedUse1904windowing, resolvedLocale, resolvedUseScientificFormat)) {
+            cache = new DataFormatterCache(resolvedUse1904windowing, resolvedLocale, resolvedUseScientificFormat);
             DATA_FORMATTER_THREAD_LOCAL.set(cache);
         }
         return cache.dataFormatter.format(data, dataFormat, dataFormatString);
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/NumberDataFormatterUtils.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/NumberDataFormatterUtils.java
@@ -40,7 +40,28 @@ public class NumberDataFormatterUtils {
     /**
      * Cache DataFormatter.
      */
-    private static final ThreadLocal<DataFormatter> DATA_FORMATTER_THREAD_LOCAL = new ThreadLocal<DataFormatter>();
+    private static final ThreadLocal<DataFormatterCache> DATA_FORMATTER_THREAD_LOCAL =
+            new ThreadLocal<DataFormatterCache>();
+
+    private static final class DataFormatterCache {
+        private final boolean use1904windowing;
+        private final Locale locale;
+        private final boolean useScientificFormat;
+        private final DataFormatter dataFormatter;
+
+        private DataFormatterCache(boolean use1904windowing, Locale locale, boolean useScientificFormat) {
+            this.use1904windowing = use1904windowing;
+            this.locale = locale;
+            this.useScientificFormat = useScientificFormat;
+            this.dataFormatter = new DataFormatter(use1904windowing, locale, useScientificFormat);
+        }
+
+        private boolean matches(boolean use1904windowing, Locale locale, boolean useScientificFormat) {
+            return this.use1904windowing == use1904windowing
+                    && this.locale.equals(locale)
+                    && this.useScientificFormat == useScientificFormat;
+        }
+    }
 
     /**
      * Format number data.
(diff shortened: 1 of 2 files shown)


alaahong (approved):
LGTM

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/NumberDataFormatterUtils.java
```diff
@@ -40,7 +40,28 @@ public class NumberDataFormatterUtils {
...
+        private final Locale locale;
+        private final boolean useScientificFormat;
+        private final DataFormatter dataFormatter;
+
+        private DataFormatterCache(boolean use1904windowing, Locale locale, boolean useScientificFormat) {
+            this.use1904windowing = use1904windowing;
+            this.locale = locale;
+            this.useScientificFormat = useScientificFormat;
+            this.dataFormatter = new DataFormatter(use1904windowing, locale, useScientificFormat);
+        }
```
pjfanning, line 57: @delei I would recommend setting this on your DataFormatter.

https://www.javadocs.dev/org.apache.poi/poi/5.5.1/org/apache/poi/ss/usermodel/DataFormatter.html#setUseCachedValuesForFormulaCells-boolean-

skytin1004: Thanks, @pjfanning. Fesod uses its own formatter and already reads cached formula results. While checking this, I found that XLS formula errors are always returned as #VALUE!, regardless of the cached error. I'll follow up with a separate fix.

PR #1070, opened 2026-09-07
Fix close unused xlsx sheet and workbook XML streams

leehaut:

Fix InputStream leaks in `XlsxSaxAnalyser` when reading XLSX files.

Hidden sheets skipped via `ignoreHiddenSheet`, sheets that are opened but never matched, and `workbook.xml` streams were not always closed. Over many files this can hold zip/package entry streams longer than needed.

## What's changed?

- Close the sheet `InputStream` immediately when the constructor skips a sheet (`ctSheet == null`), instead of dropping the handle on `continue`.
- Close every remaining stream in `sheetMap` from `execute()`'s `finally` block. Matched sheets are already closed by `parseXmlSource`; closing again is safe.
- Close `xssfReader.getWorkbookData()` with try-with-resources in `analysisUse1904WindowDate` and `analysisCtSheetMap`.
- Close failures on unused streams are logged and do not abort cleanup of other sheets.
- Keep a local `matchedSheet` instead of reassigning the loop variable.

No public API change. Read results are unchanged.

## Test plan

- [x] Added `XlsxSaxAnalyserSheetStreamCloseTest`:
  - reading only sheet 0 of a 3-sheet workbook closes unread sheet streams
  - a listener exception still closes remaining streams
  - `ignoreHiddenSheet=true` does not retain hidden-sheet streams and closes the visible ones after read
- [x] Existing `HiddenSheetsTest`, `XlsxSaxAnalyserReadOpcPackageTest`, and `DateWindowingTest` passed
- [x] `mvn -pl fesod-sheet test -Dmaven.test.skip=false -Dtest=XlsxSaxAnalyserSheetStreamCloseTest,HiddenSheetsTest,XlsxSaxAnalyserReadOpcPackageTest,DateWindowingTest`

## Checklist

- [ ] I have written the necessary doc or comment.

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/ExcelAnalyserImpl.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/ExcelAnalyserImpl.java
@@ -234,6 +234,13 @@ public class ExcelAnalyserImpl implements ExcelAnalyser {
             throwable = t;
         }
         try {
+            if (excelReadExecutor != null) {
+                excelReadExecutor.close();
+            }
+        } catch (Throwable t) {
+            throwable = t;
+        }
+        try {
             if ((readWorkbookHolder instanceof XlsxReadWorkbookHolder)
                     && ((XlsxReadWorkbookHolder) readWorkbookHolder).getOpcPackage() != null) {
                 ((XlsxReadWorkbookHolder) readWorkbookHolder).getOpcPackage().revert();
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/ExcelReadExecutor.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/analysis/ExcelReadExecutor.java
@@ -46,4 +46,12 @@ public interface ExcelReadExecutor {
      * Read the sheet.
      */
     void execute();
+
+    /**
+     * Release resources held by this executor. Invoked when the reader is finished.
+     *
+     * <p>XLSX sheet streams are opened up front and may be consumed across multiple
+     * {@link #execute()} calls, so leftover streams must not be closed until this method runs.
+     */
+    default void close() {}
 }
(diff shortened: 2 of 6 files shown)


skytin1004 (changes requested):
Hi @leehaut. The changes look good to me. Could you update the hidden-sheet test to also verify that the skipped sheet's stream is closed?

That test currently passes without the fix because it only checks that the hidden sheet is excluded from the sheet list.

nkuprins:
FYI, 

`SimpleReadCacheSelector:readCache` on `102` uses `available()` method directly on `sharedStringsTablePackagePart.getInputStream()`, so a new stream created by `getInputStream()` is not closed.

Remedy:
```java
try (InputStream inputStream = sharedStringsTablePackagePart.getInputStream()) {
    size = inputStream.available();
} catch (IOException e) {
    log.warn("Unable to get file size, default used MapCache");
    return new MapCache();
}
```

This may relate to this PR, so you might consider fixing it here as well.

leehaut:
> Hi @leehaut. The changes look good to me. Could you update the hidden-sheet test to also verify that the skipped sheet's stream is closed?
> 
> That test currently passes without the fix because it only checks that the hidden sheet is excluded from the sheet list.

Thanks, that’s a fair point. The current assertions only prove the hidden sheet is omitted from sheetList/sheetMap, which already held before this fix. I’ll add a regression that records streams closed during construction and asserts the skipped hidden sheet stream is closed.

leehaut:
> FYI,
> 
> `SimpleReadCacheSelector:readCache` on `102` uses `available()` method directly on `sharedStringsTablePackagePart.getInputStream()`, so a new stream created by `getInputStream()` is not closed.
> 
> Remedy:
> 
> ```java
> try (InputStream inputStream = sharedStringsTablePackagePart.getInputStream()) {
>     size = inputStream.available();
> } catch (IOException e) {
>     log.warn("Unable to get file size, default used MapCache");
>     return new MapCache();
> }
> ```
> 
> This may relate to this PR, so you might consider fixing it here as well.

Thanks for catching that. I’ll close that SST InputStream with try-with-resources in this PR as well, since it is the same class of leak on the XLSX read path.

skytin1004 (approved):
Thanks for the update, @leehaut. 
The new test addresses my earlier concern. LGTM!

PR #1077, opened 2026-09-07
refactor: extract format-string date detection out of DateUtils

### LINKED ISSUE #1062: [Enhancement] Extract format-string date detection out of DateUtils
### Search before asking

### Motivation

`DateUtils` mixes date parsing/formatting with format-string date detection. To improve readability and reduce complexity, I suggest refactoring it to follow SRP.

### Solution

Extract format detection into `ExcelDateFormatDetector`_(package-private)_. Move `isADateFormat`, `isADateFormatUncached`, `isInternalDateFormat`, and their regex/cache internals into a new class. 

**DateUtils keeps the same three public methods, now one-line delegations - no API change, no behaviour change.**

Example:
```java
public static boolean isADateFormat(Short formatIndex, String formatString) {
      return ExcelDateFormatDetector.isADateFormat(formatIndex, formatString);
}
```
```java
public static void removeThreadLocalCache() {
      DATE_FORMAT_THREAD_LOCAL.remove();
      DATE_TIME_FORMATTER_THREAD_LOCAL.remove();
      ExcelDateFormatDetector.removeThreadLocalCache();
}
```

### Alternatives

_No response_

### Anything else?

_No response_

nkuprins:
Closed: #1062 

As title

## What's changed?

- Extracted date format detection: moved detection logic out of `DateUtils` into a dedicated `ExcelDateFormatDetector` package-private class.
- Updated test references: adjusted `DateUtilsTest` to test date format detection via the newly extracted `ExcelDateFormatDetector`.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ExcelDateFormatDetector.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ExcelDateFormatDetector.java
@@ -26,6 +26,9 @@ import org.apache.fesod.common.util.StringUtils;
 
 /**
  * Internal helper used by {@link DateUtils} for detecting date-related spreadsheet format strings.
+ * <p>
+ * Not intended for direct use; use {@link DateUtils} as the primary entry point instead.
+ * </p>
  */
 class ExcelDateFormatDetector {
 
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ExcelDateFormatDetector.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ExcelDateFormatDetector.java
@@ -17,12 +17,6 @@
  * under the License.
  */
 
-/*
- * This file is part of the Apache Fesod (Incubating) project, which was derived from Alibaba EasyExcel.
- *
- * Copyright (C) 2018-2024 Alibaba Group Holding Ltd.
- */
-
 package org.apache.fesod.sheet.util;
 
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


skytin1004 (approved):
I reviewed this locally and ran the relevant tests, which all passed.

I confirmed that the date detection logic was moved without changes and that the existing public methods still work through DateUtils. The cache cleanup test also confirms that DateUtils.removeThreadLocalCache() clears the cache in the new class.

Since this refactoring moves the date detection logic without changing its behavior, I think the existing tests provide sufficient coverage. LGTM.

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ExcelDateFormatDetector.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.regex.Pattern;
+import org.apache.fesod.common.util.MapUtils;
+import org.apache.fesod.common.util.StringUtils;
+
+/**
+ * Extracted from {@link DateUtils}.
+ * <p>
+ * Package-private: to keep {@link DateUtils} the sole public entry point.
+ * </p>
+ */
```
bengbengbalabalabeng, line 32: Could we make the class comment focus a bit more on its responsibility? For example:

```
Internal helper used by {@link DateUtils} for detecting date-related spreadsheet format strings.
```

nkuprins: Yes, but I wanted to highlight *package-private* to avoid an accidental change to `public` by some other contributor in the future. Well, probably, *internal helper* is indeed sufficient to make it clear.

bengbengbalabalabeng: How about something like this?

```
Internal helper used by {@link DateUtils} for detecting date-related spreadsheet format strings.
<p>
Not intended for direct use; use {@link DateUtils} as the primary entry point instead.
```

bengbengbalabalabeng (approved):
LGTM.

PR #1078, opened 2026-09-08
fix: default null scientific format option to false

skytin1004:

While reviewing my formatter cache changes in #1068, I noticed a bug in the `DataFormatter` constructor when either formatting option is null.

For example, `new DataFormatter(false, Locale.US, null)` should use the default of `false` for scientific notation. Before this fix, formatting `100000000000` with the `General` format throws a `NullPointerException`. After this fix, it returns `100000000000` as expected.

Another example is `new DataFormatter(null, Locale.US, true)`. Before this fix, formatting the same number returns `100000000000`, even though scientific notation is enabled. After this fix, it returns `1E+11`.

The constructor was checking `use1904windowing` instead of `useScientificFormat` when applying the scientific-notation default. This PR changes that one condition. The examples call the constructor directly because `NumberDataFormatterUtils` already converts null options to their defaults after #1068.

## What's changed?

- Fix the condition to check `useScientificFormat`, preserving its documented default of `false`.
- Add a parameterized constructor regression test covering null, enabled, and disabled scientific-format options, including null date-windowing options.

Verification:

All 919 `fesod-sheet` tests passed on JDK 11 with the default locale set to Korean, and the build completed successfully.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/format/DataFormatter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/metadata/format/DataFormatter.java
@@ -187,7 +187,7 @@ public class DataFormatter {
             this.locale = locale;
         }
 
-        if (use1904windowing == null) {
+        if (useScientificFormat == null) {
             this.useScientificFormat = Boolean.FALSE;
         } else {
             this.useScientificFormat = useScientificFormat;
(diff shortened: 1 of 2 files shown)


bengbengbalabalabeng:
LGTM.

> **NOTE:** This change will be merged after the current release is completed. We'll follow up and push the merge forward once the release is done.

Issue #1045, opened 2026-08-26
[Bug] csv和xlsx两种模式不能共用同一个自定义Converter


luo-zhan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fesod/issues) and found nothing similar.

### Fesod version

2.0.1-incubating

### JDK version

8

### Operating system

macOS 15.2

### Steps To Reproduce

单测如下：
```java
class FesodCsvConverterKeyNativeTest {

    /** JVM 全局静态写转换器表的初始快照（本类首个测试运行前保存） */
    private static Map<Object, Object> pristineConverterMap;

    @BeforeAll
    static void snapshotGlobalConverterMap() throws Exception {
        pristineConverterMap = new HashMap<>(globalWriteConverterMap());
    }

    @BeforeEach
    void restoreGlobalConverterMap() throws Exception {
        Field field = DefaultConverterLoader.class.getDeclaredField("defaultWriteConverter");
        field.setAccessible(true);
        // 用快照的拷贝整体替换，抹掉此前测试 registerConverter 留下的全局污染
        field.set(null, new HashMap<>(pristineConverterMap));
    }

    @SuppressWarnings("unchecked")
    private static Map<Object, Object> globalWriteConverterMap() throws Exception {
        Field field = DefaultConverterLoader.class.getDeclaredField("defaultWriteConverter");
        field.setAccessible(true);
        return (Map<Object, Object>) field.get(null);
    }

    /** 内联自定义转换器：Boolean -> 是/否 */
    static class BooleanToChineseConverter implements Converter<Boolean> {
        @Override
        public Class<?> supportJavaTypeKey() {
            return Boolean.class;
        }

        @Override
        public CellDataTypeEnum supportExcelTypeKey() {
            // 返回 null 表示匹配所有 Excel 类型 —— 注册 key 为 (Boolean, null)
            return null;
        }

        @Override
        public WriteCellData<?> convertToExcelData(WriteConverterContext<Boolean> context) {
            Boolean value = context.getValue();
            if (Boolean.TRUE.equals(value)) {
                return new WriteCellData<>("是");
            }
            if (Boolean.FALSE.equals(value)) {
                return new WriteCellData<>("否");
            }
            return new WriteCellData<>("");
        }
    }

   
    @Test
    @DisplayName("xlsx 下自定义转换器正常生效：Boolean 输出「是」✅")
    void xlsxShouldApplyCustomConverter(@TempDir Path tempDir) throws Exception {
[...]

PR #1086, opened 2026-09-08
feat: honor wildcard custom converters when reading

### LINKED ISSUE #1085: [Enhancement] Custom converters registered with supportExcelTypeKey() == null never apply when reading
### Search before asking

- [x] I searched in the issues and found nothing similar.

### Fesod version

2.0.2-incubating (the code path is unchanged on current main)

### JDK version

1.8

### Operating system

Windows 10

### Minimal reproduce step

Register a custom converter whose `supportExcelTypeKey()` returns `null` (the wildcard
contract: match every cell data type), then read a file whose cells are plain strings:

```java
public static class BooleanYesNoConverter implements Converter<Boolean> {

    @Override
    public Class<?> supportJavaTypeKey() {
        return Boolean.class;
    }

    @Override
    public CellDataTypeEnum supportExcelTypeKey() {
        return null; // wildcard: match every cell type
    }

    @Override
    public Boolean convertToJavaData(ReadCellData<?> cellData, ExcelContentProperty contentProperty,
            GlobalConfiguration globalConfiguration) {
        return "yes".equalsIgnoreCase(cellData.getStringValue());
    }
}

// the file contains STRING cells "yes" and "no"
FesodSheet.read(file, BooleanReadData.class, listener)
        .registerConverter(new BooleanYesNoConverter())
        .sheet()
        .doReadSync();
```

### Current behavior

The wildcard converter is silently ignored on the read path. `ConverterUtils#convertToJavaObject`
looks the converter up with the concrete cell type key `(Boolean, STRING)`; a converter registered
under the wildcard key `(Boolean, null)` never matches, so the built-in `BooleanStringConverter`
takes over and `Boolean.valueOf("yes")` returns `false` — the row is silently read as
`flag = false`, no warning is logged.

Two more notes:

- Control case: registering the identical converter with an explicit
  `CellDataTypeEnum.STRING` key works on read — only the wildcard registration is broken.
- For types/classes without any built-in concrete-key converter, the read fails with
  `ExcelDataConvertException: Converter not found` instead.

The same registration works on the write path for xlsx, where the lookup key is
`(Boolean, null)` — so read and write disagree about the wildcard contract. See #1045 /
#1056 (write/CSV flavor of the same key mismatch, fix in progress in PR #1069); this report
covers the read path, which that PR does not touch.

### Expected behavior

Custom converters registered with `supportExcelTypeKey() == null` apply on read as well,
consistent with the wildcard contract and with the write path.

### Root cause

- Registration (`AbstractReadHolder#initConverterMap`): custom converters are registered only
  under `(supportJavaTypeKey(), supportExcelTypeKey())`, so a wildcard registration lands on
  `(Boolean, null)`.
- Lookup (`ConverterUtils#convertToJavaObject` and `ConverterUtils#convertToStringMap`): a
  single exact `converterMap.get(buildKey(clazz, cellData.getType()))` with a concrete type and
  no fallback.

### Suggested fix

Either option works; happy to implement either in a PR:

1. **Fallback lookup (recommended)**: in `ConverterUtils`, when the exact-key lookup misses and
   the cell type is non-null, retry once with the wildcard key `(clazz, null)` before failing.
   On the read side default converters are registered only under concrete keys
   (`DefaultConverterLoader#putAllConverter`), so the fallback can only hit user wildcard
   registrations: no shadowing, explicit registrations keep priority, and behavior only changes
   where reading fails or returns wrong data today.
2. **Registration expansion** (mirrors PR #1069 on the write side): additionally register
   null-key customs under every concrete cell type. Consistent with the write-side fix, but it
   introduces order-dependent overwriting between wildcard and explicit registrations.

- [x] I'm willing to submit a PR! (fix + round-trip regression tests are already prepared and
  verified: the wildcard-read case fails on current main and passes with the fix)

BigDataDZ:

Custom converters registered with `supportExcelTypeKey() == null` (the wildcard contract: "matches
every cell data type") never applied on the read path: the built-in concrete-key converter silently
took over, or reading failed with `Converter not found`. With this change, a single
`registerConverter(...)` call works for reading exactly as it already does for writing.

Closed: #1085

## What's changed?

`AbstractReadHolder` now routes custom converter registration through `registerCustomConverter`,
which — when `supportExcelTypeKey()` is `null` — additionally registers the converter under every
concrete `CellDataTypeEnum` key (all except `EMPTY`, which is filtered before lookup).

Read lookups (`ConverterUtils#convertToJavaObject`, `ConverterUtils#convertToStringMap`) use the
concrete cell type as key, so a wildcard registration at `(JavaType, null)` never matched: e.g. a
wildcard `Boolean` converter reading a `"yes"` cell produced `false` via the built-in
`BooleanStringConverter` (`Boolean.valueOf("yes")`), with no warning. This is the read-path twin of
#1045/#1056 (write/CSV flavor, addressed by PR #1069) and mirrors that PR's registration-expansion
approach, expanded to all concrete types because read lookups can arrive with any cell type.

Lookup code is untouched. Explicit-key registrations keep their priority semantics, and users
without wildcard custom converters are unaffected.

New `WildcardConverterReadTest` (round-trip): the wildcard-read case fails on current `main`
(`"yes"` read as `false`) and passes with this change; the explicit `STRING`-key case is a
regression guard. Full fesod-sheet suite: 920/920 green.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/holder/AbstractReadHolder.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/holder/AbstractReadHolder.java
@@ -142,7 +142,10 @@ public abstract class AbstractReadHolder extends AbstractHolder implements ReadH
             // A converter registered with supportExcelTypeKey() == null matches every cell type,
             // and read lookups use the concrete cell type as key (see ConverterUtils), so expand
             // each wildcard registration under every concrete key — overwriting built-in defaults
-            // while keeping the explicit registrations above in priority.
+            // while keeping the explicit registrations above in priority. Note this deliberately
+            // differs from the write side (#1069 expands to STRING only): write lookups see null
+            // targets for xlsx and force STRING only on the CSV/fill paths, while read lookups hit
+            // the actual cell type of every cell.
             for (Converter<?> converter : readBasicParameter.getCustomConverterList()) {
                 if (converter.supportExcelTypeKey() != null) {
                     continue;
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/holder/AbstractReadHolder.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/read/metadata/holder/AbstractReadHolder.java
@@ -26,7 +26,9 @@
 package org.apache.fesod.sheet.read.metadata.holder;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -34,6 +36,7 @@ import lombok.Setter;
 import org.apache.fesod.common.util.ListUtils;
 import org.apache.fesod.sheet.converters.Converter;
 import org.apache.fesod.sheet.converters.ConverterKeyBuild;
+import org.apache.fesod.sheet.converters.ConverterKeyBuild.ConverterKey;
 import org.apache.fesod.sheet.converters.DefaultConverterLoader;
 import org.apache.fesod.sheet.enums.CellDataTypeEnum;
 import org.apache.fesod.sheet.enums.HolderEnum;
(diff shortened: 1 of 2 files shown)


skytin1004 (commented):
Hi @BigDataDZ,
I tried registering a STRING converter first and a wildcard converter second. Before this change, the STRING converter handled string cells. With this change, the wildcard converter handles them instead.
Is this change intentional? If not, could you keep the STRING converter handling these cells and add a test for this registration order?

BigDataDZ:
Good catch — that behavior change was **not** intentional. Fixed in d318eea:

Custom registrations now happen in two phases — first every converter is registered at its exact
`(JavaType, CellDataType)` key, then wildcard registrations are expanded to the concrete keys
skipping already-claimed ones. Explicit registrations therefore keep priority over wildcard ones
regardless of registration order, while wildcard expansion still overrides built-in defaults
(loaded before any custom converter), so the fix for #1085 is unaffected.

Added the suggested registration-order test (`testExplicitStringKeyConverterWinsOverLaterWildcardRegistration`):
explicit STRING first + wildcard second now keeps the STRING converter handling string cells —
red before this commit, green after. Full fesod-sheet suite green.

skytin1004 (approved):
@BigDataDZ Thanks for addressing this. I verified that the explicit converter keeps precedence over a later wildcard registration, while the wildcard case still works. LGTM.

Mikkey-f (commented):
I went through the read-side lookup path to compare it with the write-side change in #1069, and the mechanism looks right. A few notes below, one of which I think is worth covering before merge.

**Verified**

- No leak risk: `AbstractReadHolder` builds a per-instance `HashMap`, and the expansion only writes to that copy — the shared default map is untouched.
- The `explicitKeys` first pass makes the "explicit wins regardless of registration order" guarantee hold (`ConverterKey` implements equals/hashCode, so `contains()` is reliable).
- Skipping `EMPTY` is safe: both lookup sites (`convertToStringMap`, `doConvertToJavaObject`) short-circuit EMPTY before reaching the map.
- In the expanded set, `ERROR` is genuinely reachable (in v03, `FormulaRecordHandler` leaves error cells as ERROR; only later formats normalise them to STRING), so the expansion is not dead code there. `RICH_TEXT_STRING` does not appear to be produced on the read side — harmless, just unused.

**Why this differs from #1069 (write side)**

Write lookups key on `(javaType, targetCellDataType)`, where xlsx targets are null (so the wildcard key matches directly) and only the CSV / fill paths force a concrete type — hence STRING is enough there. Read lookups key on the actual cell type, so honouring a wildcard means expanding to every concrete type. A one-line comment near the expansion loop would help keep this asymmetry from looking like a bug to the next reader.

**The case I would suggest addressing: the expansion also reaches the header path**

`ConverterUtils.convertToStringMap` (used by `DefaultAnalysisEventProcessor.buildHead` for header-to-field matching and by `AnalysisEventListener.invokeHeadMap`) looks up `(String.class, cellDataType)`. Because the expansion writes `(String, STRING)` and friends for a wildcard `Converter<String>`, the converter now runs on header cells as well.

Local check — bean with `@ExcelProperty("flag")`, file with header `flag` and cell `abc`, converter that upper-cases a wildcard-registered String:

- on `main`: the converter is never invoked and the value reads back as `abc` (today's no-op);
- with this change: it is invoked with "flag" (the header), the header text becomes "FLAG", name matching finds no column, and the field comes back null with no exception. With `@ExcelProperty(index = 0)` it still works, since matching then ignores the header text.

So for a wildcard String converter this turns a no-op into silently empty fields. The tests here do not catch it because they register a `Boolean` converter while the header lookup is for `String.class`.

To be clear, this is not introduced by your change — an explicit `(String, STRING)` registration already behaves this way on `main` (I checked: same silent null). But the PR widens the set of registrations that hit it, and a generic string converter is exactly the kind of thing people register wildcard-style.

I filed #1098 with the repro and root cause, and I am happy to take that fix so this PR can stay focused on the wildcard expansion — my suggestion there is that the header path should resolve against the built-in converters rather than the user-extended map. If you would rather handle it here, just say so and I will close the issue and follow your lead; either way, a test with a wildcard `String` converter would be worth adding on your side.

Worth knowing: until that header fix lands, a wildcard `Converter<String>` would give silently empty fields for name-based matching, so if you or the maintainers prefer to land the header fix first, I will move quickly on it.

BigDataDZ:
Thanks for the incredibly thorough review @Mikkey-f — the leak/ordering/EMPTY/ERROR walk-through
is exactly the kind of scrutiny this change needed.

1. **Asymmetry note added** next to the expansion loop (8dd46da), spelling out why the read side
   expands to every concrete type while #1069 only needed STRING on the write side.
2. **Wildcard `Converter<String>` test added** (`testWildcardStringConverterAppliesToDataCells`,
   index-based matching so the header path stays out of scope here): the converter applies to
   data cells as expected.
3. **Agreed on #1098** — the header path resolving against user-extended converters is a
   pre-existing flaw that this PR merely widens, and it deserves its own focused fix. It's all
   yours, much appreciated!

One sequencing note for maintainers: until #1098 lands, a wildcard `Converter<String>` remains a
hazard for name-based matching (as your analysis shows), so the merge order between this PR and
#1098 may be worth considering.

PR #1088, opened 2026-09-09
feat: add UUID string converter (#1017)

zhanghuang090:

Related: #1017

Fields declared as `java.util.UUID` currently have no built-in converter. This adds UUID ↔ STRING conversion so identifiers can be imported and exported without a custom converter.

Proposal: https://github.com/apache/fesod/issues/1017#issuecomment-5459583003

## What's changed?

- Add `UuidStringConverter`, using `UUID.fromString()` for non-empty input and `UUID.toString()` for canonical lowercase output. Empty strings are treated as missing values, consistently with blank cells.
- Register STRING reads, default writes, and explicit STRING writes in `DefaultConverterLoader`.
- Add 18 JUnit 5 cases covering lookup paths, uppercase and boundary UUIDs, malformed input, and XLSX/XLS/CSV round trips, including null fields and empty strings. Malformed non-empty cells retain the framework's `ExcelDataConvertException` with the parsing error as the cause.
- Keep the implementation compatible with Java 8 and add no dependencies.

## Validation

- Targeted UUID and existing loader tests: 22 passed (JDK 25.0.4.1, Maven 3.9.8).
- Full module regression: 60 common tests passed; 934 spreadsheet tests had 0 assertion failures and 1 existing `ClassUtilsFieldOverrideTest` cleanup error (`DirectoryNotEmptyException` while deleting a JUnit temporary directory on Windows). A focused rerun of that class plus UUID and loader tests passed all 27 cases without code changes.
- `mvn -B -ntp -pl fesod-sheet -am -Dmaven.test.skip=false -Dtest=ClassUtilsFieldOverrideTest,UuidConverterTest,DefaultConverterLoaderTest -Dsurefire.failIfNoSpecifiedTests=false package spotless:check` passed.
- New converter compiles with `javac --release 8` against the built project classes.
- `git diff --cached --check` passed.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/uuid/UuidStringConverter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/uuid/UuidStringConverter.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fesod.sheet.converters.uuid;
+
+import java.util.UUID;
+import org.apache.fesod.sheet.converters.Converter;
+import org.apache.fesod.sheet.enums.CellDataTypeEnum;
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.ReadCellData;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+
+/**
+ * Converts UUID values to canonical lowercase strings and reads strings using {@link UUID#fromString(String)}.
+ * Empty strings are treated as missing values, like blank cells.
+ */
+public class UuidStringConverter implements Converter<UUID> {
+
+    @Override
+    public Class<UUID> supportJavaTypeKey() {
+        return UUID.class;
+    }
+
+    @Override
+    public CellDataTypeEnum supportExcelTypeKey() {
+        return CellDataTypeEnum.STRING;
+    }
+
+    @Override
+    public UUID convertToJavaData(
+            ReadCellData<?> cellData, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        String value = cellData.getStringValue();
+        return value == null || value.isEmpty() ? null : UUID.fromString(value);
+    }
+
+    @Override
+    public WriteCellData<?> convertToExcelData(
+            UUID value, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        return new WriteCellData<>(value.toString());
+    }
+}
--- a/fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/UuidConverterTest.java
+++ b/fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/UuidConverterTest.java
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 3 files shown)


fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/uuid/UuidStringConverter.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.ReadCellData;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+
+/**
+ * Converts UUID values to canonical lowercase strings and reads strings using {@link UUID#fromString(String)}.
+ * Empty strings are treated as missing values, like blank cells.
+ */
+public class UuidStringConverter implements Converter<UUID> {
```
bengbengbalabalabeng, line 34: The class name should remain `UUIDStringConverter` to stay consistent with Java’s standard `java.util.UUID`. There’s no need to use the `Uuid` variant.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/uuid/UuidStringConverter.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public CellDataTypeEnum supportExcelTypeKey() {
+        return CellDataTypeEnum.STRING;
+    }
+
+    @Override
+    public UUID convertToJavaData(
+            ReadCellData<?> cellData, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        String value = cellData.getStringValue();
+        return value == null || value.isEmpty() ? null : UUID.fromString(value);
```
bengbengbalabalabeng, line 50: To make the converter more robust, you might consider using `StringUtils.isBlank(value)` to handle `null`, empty, or whitespace-only strings, and then trim the value before parsing.
Something like `UUID.fromString(value.trim())` would help avoid unexpected parsing failures caused by leading or trailing whitespace.

fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/UuidConverterTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.fesod.sheet.testkit.base.AbstractExcelTest;
+import org.apache.fesod.sheet.testkit.enums.ExcelFormat;
+import org.apache.fesod.sheet.testkit.helpers.RoundTripHelper;
+import org.apache.fesod.sheet.testkit.params.ExcelFormatSource;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class UuidConverterTest extends AbstractExcelTest {
```
nkuprins, line 50: Please add a `@Tag(Tags.ROUND_TRIP)`

fesod-sheet/src/test/java/org/apache/fesod/sheet/converter/UuidConverterTest.java
```diff
@@ -0,0 +1,152 @@
...
+                TEXT,
+                "123E4567-E89B-12D3-A456-426614174000",
+                "00000000-0000-0000-0000-000000000000",
+                "ffffffff-ffff-ffff-ffff-ffffffffffff"
+            })
+    void readsAndWritesCanonicalStrings(String input) {
+        UUID value = converter.convertToJavaData(new ReadCellData<>(input), null, configuration);
+        WriteCellData<?> cell = converter.convertToExcelData(value, null, configuration);
+        Assertions.assertEquals(CellDataTypeEnum.STRING, cell.getType());
+        Assertions.assertEquals(input.toLowerCase(java.util.Locale.ROOT), cell.getStringValue());
```
nkuprins: Please replace the full `java.util.Locale.ROOT` in both places. Use `import java.util.Locale;`.

nkuprins:
1. The test is named `UuidConverterTest`. Should we maybe rename it to uppercase `UUID...`, similar to the claim raised by bengbengbalabalabeng for `UUIDStringConverter` converter?
2. Following up on my earlier `@Tag(Tags.ROUND_TRIP)` comment: 
 Since unit tests and round-trip tests share one class, should we maybe split them? 

> One class with `@Tag(Tags.UNIT)` and another extending `AbstractExcelTest` with `@Tag(Tags.ROUND_TRIP)`

3. In `readsUppercaseAndBlankCells`, the "padded UUID" and "whitespace UUID" rows never reach the converter's own trim/blank handling: the reader trims cell text by default (`autoTrim`), so the padded row arrives as `TEXT`, and the whitespace row arrives as an empty cell or as `""`. Removing `.trim()` from the converter still passes this test. Only `trimsWhitespaceBeforeParsing` catches it. Should we drop those two rows?

nkuprins:
Nit: #1032 keeps its registration checks in `DefaultConverterLoaderTest` (`loadConvertersRegistersLocalTimeFamily`), and #1092 adds a `loadConvertersRegistersSqlTimeFamily` next to it.  If you'd like to follow that, the three `DefaultConverterLoader` assertions could move there as `loadConvertersRegistersUuidFamily`? The `supportJavaTypeKey`/`supportExcelTypeKey` assertions can stay here.

PR #1089, opened 2026-09-09
feat: support Instant string conversion

56wj:

Related: #1017

Add JDK 8-compatible support for lossless conversion between `java.time.Instant` and UTC ISO-8601 Excel string cells.

## What's changed?

- Add `InstantStringConverter` using `OffsetDateTime` for UTC-only reads and `Instant.toString()` for canonical writes.
- Register the converter for default writes, explicit string writes, and string reads.
- Reject non-UTC offsets to keep read/write string semantics symmetric.
- Add unit coverage for converter keys, UTC parsing, non-UTC rejection, nanosecond round trips, invalid input, and loader registration.
- Keep the implementation string-only because Excel date/number cells do not preserve timezone/offset semantics.

## Verification

- `mvn clean package -B -Dmaven.test.skip=false -pl fesod-common,fesod-shaded,fesod-sheet`
- `mvn -pl fesod-sheet -DskipTests spotless:check`

Results after syncing current `main` on JDK 21: 924 tests, 0 failures, 0 errors, 0 skipped. The generated Surefire reports include 7 passing `InstantStringConverterTest` cases and 5 passing `DefaultConverterLoaderTest` cases.

## Checklist

- [x] I have read the Contributor Guide.

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/instant/InstantStringConverter.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/instant/InstantStringConverter.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fesod.sheet.converters.instant;
+
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import org.apache.fesod.sheet.converters.Converter;
+import org.apache.fesod.sheet.enums.CellDataTypeEnum;
+import org.apache.fesod.sheet.metadata.GlobalConfiguration;
+import org.apache.fesod.sheet.metadata.data.ReadCellData;
+import org.apache.fesod.sheet.metadata.data.WriteCellData;
+import org.apache.fesod.sheet.metadata.property.ExcelContentProperty;
+
+/**
+ * Instant and ISO-8601 string converter.
+ */
+public class InstantStringConverter implements Converter<Instant> {
+
+    @Override
+    public Class<Instant> supportJavaTypeKey() {
+        return Instant.class;
+    }
+
+    @Override
+    public CellDataTypeEnum supportExcelTypeKey() {
+        return CellDataTypeEnum.STRING;
+    }
+
+    @Override
+    public Instant convertToJavaData(
+            ReadCellData<?> cellData, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        return OffsetDateTime.parse(cellData.getStringValue()).toInstant();
+    }
+
+    @Override
+    public WriteCellData<?> convertToExcelData(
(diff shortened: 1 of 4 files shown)


56wj:
CI on JDK 8/11 exposed a runtime compatibility difference: `Instant.parse(...)` rejected the `+08:00` test input there, although newer JDKs accepted it. I updated the read path to `OffsetDateTime.parse(...).toInstant()`, which preserves the intended ISO-8601 offset normalization and remains JDK 8 compatible. The focused converter/loader tests and Spotless check pass locally on the updated commit `87ce5b2`.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/instant/InstantStringConverter.java
```diff
@@ -0,0 +1,63 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * This file is part of the Apache Fesod (Incubating) project, which was derived from Alibaba EasyExcel.
+ *
+ * Copyright (C) 2018-2024 Alibaba Group Holding Ltd.
+ */
```
bengbengbalabalabeng: The newly added files in this PR are implemented from scratch and are not derived from Alibaba's EasyExcel. Therefore, no EasyExcel-related license header is required for these files.

56wj: Updated — I removed the EasyExcel-derived header from the new converter file and retained only the ASF license header, following `FesodSheet.java`. The focused converter/loader tests and Spotless check pass on the updated commit `3b7ae8f`.

fesod-sheet/src/main/java/org/apache/fesod/sheet/converters/instant/InstantStringConverter.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public CellDataTypeEnum supportExcelTypeKey() {
+        return CellDataTypeEnum.STRING;
+    }
+
+    @Override
+    public Instant convertToJavaData(
+            ReadCellData<?> cellData, ExcelContentProperty contentProperty, GlobalConfiguration globalConfiguration) {
+        return OffsetDateTime.parse(cellData.getStringValue()).toInstant();
+    }
```
bengbengbalabalabeng, line 57: The current implementation of `InstantStringConverter` has a semantic inconsistency between read and write behavior.

- `convertToJavaData(...)`: uses `OffsetDateTime.parse(...).toInstant()`, which accepts any offset (e.g., `+08:00`) and normalizes the value to `UTC`.
- `convertToExcelData()`: always outputs `Instant.toString()`, which is strictly `UTC`.

This means that an Excel value like `"2026-09-13T12:12:12+08:00"` will be read as `"2026-09-13T04:12:12Z"`, and writing it back will produce a different string. The round‑trip behavior is not symmetric.

Since Instant represents an absolute `UTC` timestamp, the converter should enforce that the input string is already in `UTC`. Otherwise the converter silently changes the timestamp.

I suggest adding a validation step in convertToJavaData to reject `non‑UTC` offsets. This keeps the read/write semantics consistent and avoids unexpected timezone normalization.

56wj: Updated. The read path now parses the offset explicitly and rejects any non-zero offset with `DateTimeParseException`; `Z` and `+00:00` remain valid UTC inputs, and writes stay canonical `Z`. I added a regression test for rejecting `+08:00`. The focused tests, Spotless, and the full package run all pass (924 tests) on commit `f0af0eb`.

bengbengbalabalabeng: Suggest using `getOffset().equals(ZoneOffset.UTC)` ensures that the input is truly `UTC` (Z or +00:00) and improves readability.

56wj: Updated to compare the parsed offset directly with `ZoneOffset.UTC`, as suggested. The focused tests, Spotless, and full package run all pass (924 tests) on commit `4a7c6bb`.

bengbengbalabalabeng (approved):
LGTM.

PR #1093, opened 2026-09-10
fix: preserve cached formula error values when reading XLS

skytin1004:

Preserve cached error values when reading XLS formulas. For example, a cached `1/0` result currently reads as `#VALUE!` instead of `#DIV/0!`.

Follow-up to the [discussion in apache/fesod#1068](https://github.com/apache/fesod/pull/1068#discussion_r3969328436).

## What's changed?

- Decode the cached formula error with POI's existing `ErrorEval.getText()`, as the literal-error handler already does.
- Add regression tests for seven error values in XLS and XLSX across all three read modes, including formula metadata checks.

Validation: the full `fesod-sheet` suite passed (922 tests, including six new regression cases). Spotless passed for both changed Java files.

## Checklist


skytin1004:
Hi @bengbengbalabalabeng,
when you have time, could you please take a look at this PR? It fixes cached formula error values when reading XLS, and all checks are passing. I'd appreciate your feedback. Thanks!

PR #1094, opened 2026-09-11
build: set fixed project.build.outputTimestamp for reproducible builds

alaahong:
## Summary

Sets a fixed ISO-8601 \project.build.outputTimestamp\ in \pom.xml\ so every build of a release produces identical artifacts, satisfying [reproducible-build](https://maven.org.cn/guides/mini/guide-reproducible-builds.html) requirements.

If the value is left empty, each build records a different timestamp and the resulting JARs are not reproducible.

Also documents this step in the community release guide (EN + zh-cn) so future releases remember to pin the timestamp to the release branch creation time.

## Changes
- \pom.xml\: set \project.build.outputTimestamp\ to \2026-09-10T00:00:00Z\ (fixed value) and clarify the comment.
- \website/community/release/release-version.md\: add reproducible-build step under section 3.3.1.
- \website/i18n/zh-cn/.../release-version.md\: same, in Chinese.

## Test Plan
- n/a (config + docs only).

PR #1095, opened 2026-09-11
build: drop repetitive comment on project.build.outputTimestamp

alaahong:
The project.build.outputTimestamp property kept an inline comment block
(#1094) that is not part of the Apache pom convention. This change keeps
the fixed reproducible-build timestamp but drops the repetitive comment so
the pom stays minimal and consistent with the standard boilerplate.

No functional change: only the comment lines are removed.


psxjoy (approved):
LGTM

PR #1099, opened 2026-09-12
refactor: split ClassUtils into metadata resolvers

### LINKED ISSUE #989: [Enhancement] Split ClassUtils into per-concern classes and de-duplicate the cache-location switch
### Search before asking

### Motivation

`fesod-sheet/src/main/java/org/apache/fesod/sheet/util/ClassUtils.java` is ~600 lines and holds three unrelated responsibilities in one class. It was really hard to navigate when tracing how a column or a cell format is resolved. Concretely:

**1. Column-layout resolution** - which declared fields become spreadsheet columns, and in what
order:

- `declaredFields` (L285), `doDeclaredFields` (L307), `resortField` (L394),
  `buildSortedAllFieldMap` (L449), `declaredOneField` (L472), `FieldCacheKey` (L572)
- caches: `FIELD_CACHE` (L76), `FIELD_THREAD_LOCAL` (L80)

**2. Per-field formatting resolution** - converter, date/number format and style, merged from the
head class and the row's actual runtime class:

- `declaredExcelContentProperty` (L114), `getExcelContentProperty` (L126),
  `doGetExcelContentProperty` (L149), `combineExcelContentProperty` (L168), `buildKey` (L193),
  `declaredFieldContentMap` (L197), `doDeclaredFieldContentMap` (L224), `ContentPropertyKey` (L565)
- caches: `CONTENT_CACHE` (L97), `CONTENT_THREAD_LOCAL` (L103), `CLASS_CONTENT_CACHE` (L85),
  `CLASS_CONTENT_THREAD_LOCAL` (L91)

**3. `getAllInterfaces`** (L530/L547) -  a generic reflection helper unrelated to either of the
above. It is a trimmed copy of `org.apache.commons.lang3.ClassUtils.getAllInterfaces`, and its
only caller is `FieldUtils.getField` (`FieldUtils.java:165`).

On top of that, concerns (1) and (2) each cache their result per `CacheLocationEnum`
(`THREAD_LOCAL` / `MEMORY` / `NONE`), and the switch that dispatches on the cache location is
repeated three times - L128-146, L202-221 and L286-304. 
The three differ only in the map and key types. The branch order, the lazy `ThreadLocal` init block and the 
`default: throw new UnsupportedOperationException("unsupported enum")` are identical.

### Solution

Split the file four ways within `org.apache.fesod.sheet.util`, as a pure move - no logic changes:

1. **`FieldCacheUtils`** - column-layout resolution: `declaredFields`, `doDeclaredFields`, `resortField`, `buildSortedAllFieldMap`, `declaredOneField`, `FieldCacheKey`, plus its two cache fields.
2. **`ExcelContentPropertyUtils`** - per-field formatting resolution: `declaredExcelContentProperty`, `getExcelContentProperty`, `doGetExcelContentProperty`, `combineExcelContentProperty`, `buildKey`, `declaredFieldContentMap`, `doDeclaredFieldContentMap`, `ContentPropertyKey`, plus its four cache fields.
3. **`CacheLocationUtils`** - one generic method replacing the three switches:

   ```java
   static <K, V> V computeIfAbsent(
           CacheLocationEnum cacheLocation,
           ThreadLocal<Map<K, V>> threadLocalCache,
           Map<K, V> memoryCache,
           Supplier<K> keySupplier,
           Supplier<V> valueSupplier) { ... }
   ```

4. **`ClassUtils`** - trimmed to `getAllInterfaces` plus `removeThreadLocalCache`, kept as a public entry point so its two callers don't change.

The two clusters are almost disjoint: neither calls into the other and they share no private helper. The one exception is `removeThreadLocalCache` (L591), which clears all three `ThreadLocal`s and so spans both clusters. It has two production callers (`ExcelAnalyserImpl.java:298`, `WriteContextImpl.java:567`), so it stays on `ClassUtils` as a delegate to a package-private clear method on each of the two new classes - that keeps both call sites untouched. Apart from that one method, the partition needs no duplication and no new coupling.

### Alternatives

_No response_

### Anything else?

**One caveat.** `ClassUtils` is a public utility class - some things to be moved are `public static`. If refactoring takes place, not sure if `@Deprecated` would be needed instead of just removing.

nkuprins:
Closed: #989 

As title

## What's changed?

- Extracted column-layout resolution (`declaredFields` and its helpers) into a package-private `SheetHeadFieldResolver`.
- Extracted per-field formatting resolution (`declaredExcelContentProperty` and its helpers) into a package-private `SheetContentPropertyResolver`.
- Replaced the three switches with a `MetadataCacheStrategy` per `CacheLocationEnum` constant (`InMemoryCache`, `ThreadLocalCache`, `NoOpCache`), each owning its own storage.  `MetadataCaches` composes the three for one kind of metadata and picks the tier from the `ConfigurationHolder`, so `CacheLocationEnum` is now referenced in one class instead of three switches. `InMemoryCache` and `ThreadLocalCache` also accept a map or map factory, to allow map types other than the defaults later.
- `ClassUtils` keeps `getAllInterfaces`, both key classes, and every public entry point, delegating to the resolvers.
- Deprecated `FIELD_CACHE`, `CLASS_CONTENT_CACHE` and `CONTENT_CACHE` in favour of `getFieldCache()` / `getClassContentCache()` / `getContentCache()` (unmodifiable, live views; the cached values are shared and must not be modified) and a new `removeInMemoryCache()`. The fields still alias the same live, mutable maps, so downstream reads and writes behave exactly as today - only a deprecation warning is added.
- Tests: `CacheDataTest` no longer reads a private `ThreadLocal` field of `ClassUtils` and checks that the thread-local cache is cleared after a read through `ClassUtils` itself; `ClassUtilsTest` covers the three views; new `MetadataCacheStrategyTest` checks that `ThreadLocalCache.clear()` detaches the map from the thread.

## Checklist

### DIFF
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
@@ -27,7 +27,11 @@ import org.apache.fesod.common.util.MapUtils;
 import org.apache.fesod.sheet.enums.CacheLocationEnum;
 
 /**
- * Where one kind of metadata is cached, one implementation per {@link CacheLocationEnum} constant.
+ * Internal helper used by {@link SheetHeadFieldResolver} and {@link SheetContentPropertyResolver} for
+ * caching resolved metadata, one implementation per {@link CacheLocationEnum} constant.
+ * <p>
+ * Not intended for direct use; use {@link ClassUtils} as the primary entry point instead.
+ * </p>
  *
  * @param <K> the cache key
  * @param <V> the cached metadata
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
@@ -39,6 +39,13 @@ interface MetadataCacheStrategy<K, V> {
     V get(K key, Function<K, V> mappingFunction);
 
     /**
+     * Drops the cached entries. {@code ThreadLocal}-backed implementations must detach the entry from
+     * the thread ({@link ThreadLocal#remove()}) rather than merely empty the map they hold, otherwise
+     * pooled threads keep the entry forever.
+     */
+    void clear();
+
+    /**
      * The cache will not be cleared unless the app is stopped.
      */
     class InMemoryCache<K, V> implements MetadataCacheStrategy<K, V> {
@@ -52,6 +59,11 @@ interface MetadataCacheStrategy<K, V> {
         @Override
         public V get(K key, Function<K, V> mappingFunction) {
             return cache.computeIfAbsent(key, mappingFunction);
+        }
+
+        @Override
+        public void clear() {
+            cache.clear();
         }
     }
 
@@ -76,6 +88,11 @@ interface MetadataCacheStrategy<K, V> {
             }
             return cacheMap.computeIfAbsent(key, mappingFunction);
         }
+
+        @Override
+        public void clear() {
+            cache.remove();
+        }
     }
 
     /**
@@ -86,6 +103,11 @@ interface MetadataCacheStrategy<K, V> {
         @Override
         public V get(K key, Function<K, V> mappingFunction) {
             return mappingFunction.apply(key);
+        }
+
+        @Override
+        public void clear() {
+            // nothing is cached
         }
     }
 }
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
@@ -46,7 +46,7 @@ interface MetadataCacheStrategy<K, V> {
     static <K, V> Map<CacheLocationEnum, MetadataCacheStrategy<K, V>> byLocation(
             Map<K, V> memoryCache, ThreadLocal<Map<K, V>> threadLocalCache) {
         Map<CacheLocationEnum, MetadataCacheStrategy<K, V>> strategies = new EnumMap<>(CacheLocationEnum.class);
-        strategies.put(CacheLocationEnum.MEMORY, new MemoryCache<>(memoryCache));
+        strategies.put(CacheLocationEnum.MEMORY, new InMemoryCache<>(memoryCache));
         strategies.put(CacheLocationEnum.THREAD_LOCAL, new ThreadLocalCache<>(threadLocalCache));
         strategies.put(CacheLocationEnum.NONE, new NoOpCache<>());
         return Collections.unmodifiableMap(strategies);
@@ -68,11 +68,11 @@ interface MetadataCacheStrategy<K, V> {
     /**
      * The cache will not be cleared unless the app is stopped.
      */
-    class MemoryCache<K, V> implements MetadataCacheStrategy<K, V> {
+    class InMemoryCache<K, V> implements MetadataCacheStrategy<K, V> {
 
         private final Map<K, V> cache;
 
-        MemoryCache(Map<K, V> cache) {
+        InMemoryCache(Map<K, V> cache) {
             this.cache = cache;
         }
 
--- a/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
+++ b/fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
@@ -19,16 +19,14 @@
 
 package org.apache.fesod.sheet.util;
 
-import java.util.Collections;
-import java.util.EnumMap;
 import java.util.Map;
 import java.util.function.Function;
 import org.apache.fesod.common.util.MapUtils;
 import org.apache.fesod.sheet.enums.CacheLocationEnum;
 
 /**
- * Internal helper used by {@link SheetHeadFieldResolver} and {@link SheetContentPropertyResolver} for
- * caching resolved metadata, one implementation per {@link CacheLocationEnum} constant.
+ * Internal helper used by {@link MetadataCaches} for caching resolved metadata, one implementation per
+ * {@link CacheLocationEnum} constant.
  * <p>
  * Not intended for direct use; use {@link ClassUtils} as the primary entry point instead.
  * </p>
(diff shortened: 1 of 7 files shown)


nkuprins:
@bengbengbalabalabeng PTAL, when it's convenient for you 🙏🏻

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
```diff
@@ -0,0 +1,159 @@
...
+        public void clear() {
+            cache.clear();
+        }
+
+        /**
+         * Read-only view of the cached entries.
+         */
+        Map<K, V> view() {
+            return view;
+        }
```
bengbengbalabalabeng, line 76: I've noticed that `view()` appears to always be empty. I'd like to confirm whether this is the expected behavior.

nkuprins: Yes, that's expected. `view()` only reflects the `MEMORY` tier, and the default `filedCacheLocation` is `THREAD_LOCAL`, so with default settings it stays empty. Look at `ClassUtils` in the current `main`. All 3 fields are filled only in the `case MEMORY:` branches (`ClassUtils.java:138, :213, :290`)

When they fill up:
```java
FesodSheet.write(file, Demo.class)
        .filedCacheLocation(CacheLocationEnum.MEMORY)
        ...
```

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
```diff
@@ -0,0 +1,159 @@
...
+         */
+        ConcurrentHashMap<K, V> backingMap() {
+            if (!(cache instanceof ConcurrentHashMap)) {
+                throw new IllegalStateException(
+                        "The deprecated ClassUtils cache fields require a ConcurrentHashMap, but got "
+                                + cache.getClass().getName());
+            }
+            return (ConcurrentHashMap<K, V>) cache;
+        }
+    }
```
bengbengbalabalabeng, line 84: Since `backingMap()` strongly depends on `ConcurrentHashMap`, I'd suggest removing the parameterized constructor for now and initializing it directly in the no-args constructor. This ensures the strong consistency and safety of the current API. Later on, once the technical debt in `ClassUtils` is cleared and we actually need to introduce other `Map` implementations, we can always re-introduce the parameterized constructor."

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
```diff
@@ -0,0 +1,159 @@
...
+
+        @Override
+        public V get(K key, Function<K, V> mappingFunction) {
+            Map<K, V> cacheMap = cache.get();
+            if (cacheMap == null) {
+                cacheMap = mapFactory.get();
+                cache.set(cacheMap);
+            }
+            return cacheMap.computeIfAbsent(key, mappingFunction);
+        }
```
bengbengbalabalabeng, line 109: It might be worth considering `ThreadLocal.withInitial(mapFactory)` here to simplify the implementation.

This would remove the redundant `mapFactory` field and reduce the logic in `get(...)` to a single `computeIfAbsent` call (`cache.get().computeIfAbsent(key, mappingFunction)`). Also explicitly perform a fail-fast non-null check on `mapFactory` at the constructor entry.

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/SheetContentPropertyResolver.java
```diff
@@ -0,0 +1,225 @@
...
+        return new ContentPropertyKey(clazz, headClass, fieldName);
+    }
+
+    private static Map<String, ExcelContentProperty> resolveClassContent(
+            Class<?> clazz, ConfigurationHolder configurationHolder) {
+        if (clazz == null) {
+            return null;
+        }
+        return CLASS_CONTENT_CACHES.get(configurationHolder, clazz, key -> doResolveClassContent(clazz));
+    }
```
bengbengbalabalabeng: The name `resolveClassContent` might be slightly misleading. Since it resolves the `ExcelContentProperty` mapping for all fields in the class, would `resolveContentProperties` (or `resolveContentPropertyMap`) be more precise?

nkuprins: Yes, `resolveContentPropertyMap` will make it precise

fesod-sheet/src/main/java/org/apache/fesod/sheet/util/MetadataCacheStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+        /**
+         * The live map, only for the deprecated public cache fields on {@link ClassUtils}; remove with them.
+         */
+        ConcurrentHashMap<K, V> backingMap() {
+            return cache;
+        }
+    }
+
+    /**
+     * The cache will be stored in {@code ThreadLocal}, and will be cleared when the excel read and
```
bengbengbalabalabeng: It may be more appropriate to use "spreadsheet" instead of "excel" to keep the terminology generic.

fesod-sheet/src/test/java/org/apache/fesod/sheet/util/MetadataCacheStrategyTest.java
```diff
@@ -0,0 +1,52 @@
...
+                    return new HashMap<>();
+                });
+
+        cache.get("key", key -> "value");
+        cache.get("key", key -> "value");
+        Assertions.assertEquals(1, createdMaps.get());
+
+        cache.clear();
+        cache.get("key", key -> "value");
+        Assertions.assertEquals(2, createdMaps.get());
```
bengbengbalabalabeng, line 50: Could we wrap the test logic in a `try ... finally { cache.clear(); }` block?

At the end of the test, `cache.get("key", ...)` is called again after `clear()`, which re-attaches a new map to the current thread. If the test runner reuses threads across test cases (or if an assertion fails midway), this may leave stale `ThreadLocal` data and risk polluting other tests.

nkuprins: But this test constructs its own `ThreadLocalCache` instance in the method body, so `ThreadLocal` in this test isn't shared with anything, unlike `ClassUtilsTest `that needs its `@AfterEach` because it exercises the static caches on `ClassUtils`. Could you clarify this point, please?

bengbengbalabalabeng: You're right. I was mainly thinking of the general defensive pattern around `ThreadLocal`, but overlooked that it's an isolated local instance here. No risk of polluting other tests indeed.

bengbengbalabalabeng:
LGTM.

bengbengbalabalabeng:
> LGTM.

As `ClassUtils` is part of the core infrastructure, it's safer to get another committer/maintainer's review and approval before merging.

skytin1004 (approved):
Reviewed the latest changes. LGTM.

PR #1101, opened 2026-09-14
 fix: reject out-of-range integral cell values 

### LINKED ISSUE #1100: [Bug] Integer, Long, Short and Byte fields return wrong values when a cell is out of range
### Search before asking

### Fesod version

main

### JDK version

any

### Operating system

Linux

### Steps To Reproduce

```java
@Getter
@Setter
public static class Row {
    @ExcelProperty(index = 0)
    private Integer phone;
}

File file = new File("overflow.xlsx");
FesodSheet.write(file).sheet()
        .doWrite(Collections.singletonList(Collections.singletonList(13800138000L)));

List<Row> rows = FesodSheet.read(file).head(Row.class).headRowNumber(0).sheet().doReadSync();
System.out.println(rows.get(0).getPhone());
```

### Current Behavior

The read succeeds and prints `915236112` (that is `13800138000 mod 2^32`). No exception, no warning.

Other integer types wrap the same way. The wrong value is returned in XLSX, XLS and CSV, for number cells and text cells alike:

| Field type | Cell value | Read back |
|---|---|---|
| `Integer` | `13800138000` | `915236112` |
| `Short` | `40000` | `-25536` |
| `Byte` | `"300"` | `44` |
| `Long` | `Long.MAX_VALUE` (XLSX/XLS, stored as a double) | `-9223372036854771616` |

### Expected Behavior

An out-of-range value should throw `ArithmeticException`.

### Anything else?

`intValueExact()` alone isn't enough because it also rejects any fraction.
The fix is to check the value against the type's range with `BigDecimal.compareTo` before narrowing and throw in the worst case.

nkuprins:
Closed: #1100 

As title.

## What's changed?

- Integer converters and `NumberUtils.parseByte/parseShort/parseInteger/parseLong` now range-check the value before narrowing, so out-of-range values throw instead of wrapping around.
- The integer `parse*` methods now call `parseBigDecimal`, which already has the `hasFormat` branch.
- Added out-of-range tests to `ConverterTest` and `NumberUtilsTest`.

## Checklist


skytin1004 (approved):
Thanks for the fix. I reproduced the issue on main and verified the fix for XLSX, XLS, and CSV. LGTM.

